PR #11, opened 2019-02-17
spindle_on speed update

LettyYeti:
Updated from M3 S12000 to M3 S25000 under spindle_on in the router_machine.py


Neex101:
I love github

PR #58, opened 2019-03-20
Filechooser fix

LettyYeti:
Filechooser screen refresh updated; just needed a different directory to look for ftp files.

Have not checked that it definitely finds files ftp'd across to it. 

Also chucked a "reboot" into the Get SW Update in dev.


LettyYeti:
Filechooser screen refresh updated; just needed a different directory to look for ftp files.

Have not checked that it definitely finds files ftp'd across to it.

Also chucked a "reboot" into the Get SW Update in dev.

PR #68, opened 2019-03-26
Add pausing

LettyYeti:


LettyYeti:
Added pausing functionality to go screen - just buttons. 

Now need to do same for error screens that pop up during streaming.

PR #80, opened 2019-03-28
Homing screens - they are a Thing!!

LettyYeti:
WARNING: Error screens at present are suppressed during homing!!


Neex101:
Don't know if this is relevant to our context, and don't fully understand, but there's a warning in the docs about `.unschedule`: 

"Calling CyClockBase.unschedule() on the original callback is highly discouraged because it’s significantly slower than when using the event."

https://kivy.org/doc/stable/api-kivy.clock.html#threading-and-callback-order

LettyYeti:
Yup that's fine, used it in the recommended manner ;)

PR #116, opened 2019-04-06
Magic square axis fix

Neex101:
Changed how we handle the two different homing cases.

* Homing screen now checks serial_comm's `is_sequentially_streaming` flag to know whether it's still homing or not. 
* Squaring routing now uses the `G4 P<x>` grbl delay command to force a 'blocking ok' which guarantees that that machine is idle before issuing $ commands. This replaces a ton of fragile code which was checking for states etc.
* Serial_comms now has a function cancel sequential streaming as a precaution

### DIFF
--- a/src/asmcnc/comms/router_machine.py
+++ b/src/asmcnc/comms/router_machine.py
@@ -99,39 +99,8 @@ class RouterMachine(object):
 
     def home_all(self):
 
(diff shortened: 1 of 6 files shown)


LettyYeti:
might want to add in an
 if m.state() != Idle:
    some error message

before homing starts

PR #150, opened 2019-05-19
Ask to check gcode before go

Neex101:
Think this is set to merge to master,-dev?
Removed compulsory req to grbl-check GCode.


LettyYeti:
- While new screen for file checking exists, it's currently not added to the screen manager or called anywhere so shouldn't create any unnecessary slowdown :)

PR #267, opened 2020-03-17
Remove obsolete buttons from map (#257)

Neex101:
In Pro, from virtual map tab, removes obsolete view buttons. Adds whitespace between "set" and "go" functions to reduce risk of mis-commands.

### DIFF
--- a/src/asmcnc/skavaUI/widget_virtual_bed_control.py
+++ b/src/asmcnc/skavaUI/widget_virtual_bed_control.py
@@ -82,6 +82,20 @@ Builder.load_string("""
                         size: self.parent.width, self.parent.height
                         allow_stretch: True       
         
+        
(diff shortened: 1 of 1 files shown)


src/asmcnc/skavaUI/widget_virtual_bed_control.py
```diff
@@ -136,6 +82,20 @@
...
+            pos: self.parent.pos   
+#             padding: 5
+#             spacing: 5
+#             orientation: "horizontal"
+#             canvas:
+#                 Color: 
+#                     rgba: hex('FFFFFFFF')
+#                 RoundedRectangle: 
+#                     size: self.size
+#                     pos: self.pos
```
Neex101: This is a stub for future button additions if needed

PR #268, opened 2020-03-17
Added "M06" to blacklist of gcode lines

### LINKED ISSUE #220: add any lines with an "M06" to the streaming blacklist

Neex101:
Lines containing this M06 will get dropped from the job_file.
(Previously "M6" *was* caught, but some protocols export this as M06, so the cmd was sneaking through).

Closes #220


LettyYeti:
Also added same line strip to shapecutter

PR #269, opened 2020-03-17
Enforce minimum rpms for M3 spindle cmd

### LINKED ISSUE #235: File check - alert user if spindle speed is too low
SmartBench doesn't recognise any speed setting below 3,500rpm.

Neex101:
Typical line of gcode: `M3 S1000`
`M3` switches on spindle
`S` defines rpm
Auto-ammends gcode if `S` param is lower than `minimum_spindle_rpm`

Closes #235


Neex101:
Oooh. Hadn't considered this case before. 
(Normally the syntax is `M3 S5000`, but this looks like the file you use had `S5000 M3` - grbl clearly operates both
Out of interest where did you get this file?

Neex101:
@Lettie this should be fixed now to handle either M3 protocol. Passed both ok. Ready for you to retest and merge PR...

PR #283, opened 2020-03-18
Drop img preview container in filechoosers and dangerous usb 'clean on move'

### LINKED ISSUE #282: Drop img preview container in filechoosers
This container isn't useful anymore and is reducing the ability to have long filenames

Neex101:
Img preview was defunct anyway, so dropped for both local and usb filechoosers.
Now bigger space for longer filenames.
Also dropped switch on the usb filechooser which was designed to delete the file from the usb stick after copying it to the console. It was dangerous (caused crashes) as wasn't used enough to fix.

Closes #282


Neex101:
@LettyYeti this is implemented and ready for your review.
But be aware that kivy has a bug which means that the last file never gets wiped off the filechooser screen despite refreshing the widget. I've battled with this for days before, so we will have to accept it for now.

Issue #251, opened 2020-02-18
Bounding box TRACE doesn't know difference between datum and starting point


LettyYeti:
Using the trace command on the move screen (i.e. the box) traces the bounding box of where the job is cut from the start of the cut, but starts tracing from the datum - this translates the whole bounding box from cut start to datum

LettyYeti:
Job envelope function really needs a full re-write

PR #290, opened 2020-03-20
Fix for #251: Added go to datum commands on bounding box buttons in x-y widgets

LettyYeti:


Neex101:
@LettyYeti we need to chat on this - there's a conflict of intents here

PR #311, opened 2020-03-22
Changed on_releases to on_presses...

LettyYeti:
unless something like cancel jog needed


LettyYeti:
@Neex101 please review changes with new eyes in case of dumb mistakes

PR #371, opened 2020-04-15
Output git messages

LettyYeti:
- SW updat failure recognition and repair (recloning repo) capabilities added


LettyYeti:
Need to 
- still enable SW repair if it can't get versions bc it's corrupted
- check all pop up formatting

Issue #36, opened 2019-02-23
a few bloated files in src/asmcnc/skavaUI/img/


Neex101:
Unforgivable bloatation

LettyYeti:
just make these smaller (file size)

PR #385, opened 2020-04-17
No more bloated files in src

Neex101:
image

Fat files replaced with bare minium pngs - should make for slightly better load times and improve GPU capacity.
Note this probs won't improve download time since git repos save everything into the repo forever - in the above file tree space visualisation, that big blue blob is our history.... Extraction is possible, but a project which requires git gloves.


Neex101:
Closes #259 , #36

PR #420, opened 2020-07-02
Hide spindle overload for non-pro users

LettyYeti:
- Currently don't have product codes, but all logic is in place for when we know what they are
- Added password protection into dev screen for "toggle dev mode" 
- If dev mode is on, then spindle overload widget is displayed regardless of product code :)


LettyYeti:
Checks for a serial number + product code that ends in "03", indicating precision pro. If it finds this, it shows spindle overload. 
Added password protection into dev screen for "toggle dev mode"
- If dev mode is on, then spindle overload widget is displayed regardless of product code :)

PR #504, opened 2020-11-12
Laser time

LettyYeti:
Moved laser off into bootup sequence, changed time to command, and set default LED to yellow so that when laser off called on startup it doesn't override the LED


LettyYeti:
changed order of grbl commands as well

PR #795, opened 2022-02-01
Temperatures and voltages into serial comms

LettyYeti:
Also contains unit test for temps & voltages with mock serial object (needs mock and serial_mock packages to run)

### DIFF
--- a/src/asmcnc/comms/serial_connection.py
+++ b/src/asmcnc/comms/serial_connection.py
@@ -876,7 +876,7 @@ class SerialConnection(object):
                     except IndexError:
                         pass
 
-                    except Exception as E:
+                    except:
                         log("ERROR status parse: Temperature invalid: " + message)
                         return
 
@@ -898,8 +898,7 @@ class SerialConnection(object):
                     self.PSU_mV = float(voltages[2])
                     self.spindle_speed_monitor_mV = float(voltages[3])
 
-                else:
-                    continue
+                # end of for loop
 
             if self.VERBOSE_STATUS: print (self.m_state, self.m_x, self.m_y, self.m_z,
                                            self.serial_blocks_available, self.serial_chars_available)
--- a/src/asmcnc/comms/serial_connection.py
+++ b/src/asmcnc/comms/serial_connection.py
@@ -855,6 +855,49 @@ class SerialConnection(object):
                     self.feed_rate = feed_speed[0]
                     self.spindle_speed = feed_speed[1]
 
+                # TEMPERATURES
+                elif part.startswith('TC:'):
+                    temps = part[3:].split(',')
+
+                    try: 
+                        float(temps[0])
+                        float(temps[1])
+                    except:
+                        log("ERROR status parse: Temperature invalid: " + message)
+                        return
+
+                    self.pcb_temp = float(temps[0])
+                    self.motor_driver_temp = float(temps[1])
+
+                    try:
+                        float(temps[2])
+                        self.transistor_heatsink_temp = float(temps[2])
+
+                    except IndexError:
+                        pass
+
+                    except Exception as E:
+                        log("ERROR status parse: Temperature invalid: " + message)
+                        return
+
+                # VOLTAGES
+                elif part.startswith('V:'):
+                    voltages = part[2:].split(',')
+                    try: 
+                        float(voltages[0])
+                        float(voltages[1])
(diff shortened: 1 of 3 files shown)


src/asmcnc/comms/serial_connection.py
```diff
@@ -845,6 +855,49 @@ def process_grbl_push(self, message):
...
+
+                    self.pcb_temp = float(temps[0])
+                    self.motor_driver_temp = float(temps[1])
+
+                    try:
+                        float(temps[2])
+                        self.transistor_heatsink_temp = float(temps[2])
+
+                    except IndexError:
+                        pass
```
archiepeterjarvis, line 877: intentionally passing?

src/asmcnc/comms/serial_connection.py
```diff
@@ -845,6 +855,49 @@ def process_grbl_push(self, message):
...
+
+                    except Exception as E:
```
archiepeterjarvis, line 879: specify exception type?

archiepeterjarvis: plus no reference to E is ever made as far as I can tell

src/asmcnc/comms/serial_connection.py
```diff
@@ -845,6 +855,49 @@ def process_grbl_push(self, message):
...
+                        log("ERROR status parse: Voltage invalid: " + message)
+                        return
+
+                    self.microcontroller_mV = float(voltages[0])
+                    self.LED_mV = float(voltages[1])
+                    self.PSU_mV = float(voltages[2])
+                    self.spindle_speed_monitor_mV = float(voltages[3])
+
                 else:
                     continue
```
archiepeterjarvis, line 902: is this redundant?

test/comms_serial_connection_tests/tempsvoltagestest.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+    def tearDown(self):
+      self.serial_module.__del__()
+
+
+    def test_does_serial_think_its_connected(self):
+        """Test that serial module thinks it is connected"""
+        self.status_and_PCB_constructor()
+        assert self.serial_module.is_connected() == True, 'not connected'
```
archiepeterjarvis, line 180: for each assertion, could be changed to

`assert self.serial_module.is_connected(), 'not connected'`

src/asmcnc/comms/serial_connection.py
archiepeterjarvis: (not certain on context but may be worth logging regardless)

LettyYeti: yes - so in the FW versions 1.3.6 - now, they've only reported 2 temps instead of 3, so this will try to retrieve the 3rd temp but if it isn't there then it'll just keep the code going :) 

I'm fairly sure this is the most "pythonic" way to handle this case

src/asmcnc/comms/serial_connection.py
LettyYeti: good catch - was using it for debugging, will fix 😁

src/asmcnc/comms/serial_connection.py
LettyYeti: not sure what you mean? which bit?

test/comms_serial_connection_tests/tempsvoltagestest.py
LettyYeti: oooh yisss, much tidier

src/asmcnc/comms/serial_connection.py
archiepeterjarvis: the else: which only contains a continue - is it necessary? code would continue regardless

LettyYeti: ooh, no I don't think it is - just was always in there so I left it 😂

src/asmcnc/comms/serial_connection.py
archiepeterjarvis: Yep that makes sense, for completeness sake is there any reason why it would help to see a log of this reason? For example if a machine hasn't been updated?

LettyYeti: Nah, would just be extra noise at the mo - not all FW can go on all PCBs anyway, and currently we don't update FW in the field

PR #798, opened 2022-02-03
Parses 4-value Ld: and single-value Ld: messages differently

LettyYeti:
4-value Ld: messages are currently just recorded in serial module

### DIFF
--- a/src/asmcnc/comms/router_machine.py
+++ b/src/asmcnc/comms/router_machine.py
@@ -1204,6 +1204,7 @@ class RouterMachine(object):
 
 # SPEED AND FEED GETTERS
     def feed_rate(self): return int(self.s.feed_rate)
+
     def spindle_speed(self): 
         if self.spindle_voltage == 110: 
             # if not self.spindle_digital or not self.fw_can_operate_digital_spindle(): # this is only relevant much later on
@@ -1211,7 +1212,12 @@ class RouterMachine(object):
             return int(converted_speed)
         else: 
             return int(self.s.spindle_speed)
-    def spindle_load(self): return int(self.s.spindle_load_voltage)
+
+    def spindle_load(self): 
+        try:
+            return int(self.s.spindle_load_voltage)
+        except:
+            return ''
 
 # POSITIONAL SETTERS
 
--- a/src/asmcnc/comms/serial_connection.py
+++ b/src/asmcnc/comms/serial_connection.py
@@ -637,7 +637,15 @@ class SerialConnection(object):
     # Feeds and speeds
     spindle_speed = '0.0'
     feed_rate = '0.0'
-    spindle_load_voltage = 0.0
+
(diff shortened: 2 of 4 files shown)


src/asmcnc/comms/router_machine.py
```diff
@@ -1204,14 +1204,20 @@ def g28_z(self): return float(self.s.g28_z)
...
 # SPEED AND FEED GETTERS
     def feed_rate(self): return int(self.s.feed_rate)
+
     def spindle_speed(self): 
         if self.spindle_voltage == 110: 
             # if not self.spindle_digital or not self.fw_can_operate_digital_spindle(): # this is only relevant much later on
             converted_speed = self.convert_from_230_to_110(self.s.spindle_speed)
             return int(converted_speed)
         else: 
             return int(self.s.spindle_speed)
```
archiepeterjarvis, line 1214: are explicit casts required?

src/asmcnc/comms/router_machine.py
```diff
@@ -1204,14 +1204,20 @@ def g28_z(self): return float(self.s.g28_z)
...
-    def spindle_load(self): return int(self.s.spindle_load_voltage)
+
+    def spindle_load(self): 
+        try:
+            return int(self.s.spindle_load_voltage)
+        except:
+            return ''
```
archiepeterjarvis, line 1220: potentially unsafe? returning strict int in good case otherwise returning a string

src/asmcnc/comms/serial_connection.py
```diff
@@ -825,37 +833,70 @@ def process_grbl_push(self, message):
...
+
+                    if ',' in spindle_feedback:
+
+                        digital_spindle_feedback = spindle_feedback.split(',')
+
+                        try: 
+                            int(digital_spindle_feedback[0])
+                            int(digital_spindle_feedback[1])
+                            int(digital_spindle_feedback[2])
+                            int(digital_spindle_feedback[3])
```
archiepeterjarvis, line 847: check for split length instead of relying on try?

archiepeterjarvis: this actually makes no difference :)

src/asmcnc/comms/router_machine.py
LettyYeti: yep, it's complicated - will explain another time!

src/asmcnc/comms/router_machine.py
LettyYeti: oooh good point... it's default value in serial is None (so I was trying to protect against it passing that back) but actually maybe that's better than an int/str situation. 

the plus side is that at the moment, it isn't actually used by anything else in the code 😅

LettyYeti: tempted to leave it for now until there's an actual 'use case' for it

PR #832, opened 2022-03-25
Serial number validation

archiepeterjarvis:


src/asmcnc/production/lowerbeam_calibration_jig/lowerbeam_calibration_4.py
```diff
@@ -65,7 +70,22 @@ def __init__(self, **kwargs):
...
+
+        validated = self.validate_serial_number(serial_number)
+
+        if not validated:
+            self.error_label.text = 'Serial number invalid'
+            return
+
         try:
             self.send_calibration_payload(TMC_Y1)
             self.send_calibration_payload(TMC_Y2)
```
LettyYeti, line 91: should the set_serial_no() functions be set_serial_no(serial_number) instead of set_serial_no(self.serial_no_input.text) ?

PR #836, opened 2022-04-06
Complete revamp of calibration process

LettyYeti:
- Uses database v0.4, which saves all datapoints in individual rows as per B's request
- Requires serial number entry for all sub assemblies
- Collects statistics (peak and average load values) as well as raw status information
- Unweighted/weighted load tests refactored so that they now run all the way up and down each axis once
- User interface modified to show peaks for each stage of process in overnight test & load test screens
- Peak values are changed back to peak positive values, rather than absolute peaks
- Added recalibration stage into final test (after 6 hour wear-in)
- Added in post-recalibration check, and then another run
- Change using Ansible to make serial connection happen over ttyAMA0 port (instead of ttyS0) - can't roll back from this version to earlier than v2!!

### DIFF
--- a/src/asmcnc/apps/systemTools_app/screens/screen_factory_settings.py
+++ b/src/asmcnc/apps/systemTools_app/screens/screen_factory_settings.py
@@ -830,7 +830,7 @@ class FactorySettingsScreen(Screen):
     def enter_serial_number_screen(self):
         # if self.calibration_db.conn != None:
         if not self.systemtools_sm.sm.has_screen('serial_input_screen'):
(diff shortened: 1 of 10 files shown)


src/asmcnc/apps/systemTools_app/screens/screen_factory_settings.py
```diff
@@ -862,6 +862,15 @@ def enter_overnight_test(self):
         else:
             popup_info.PopupError(self.systemtools_sm, self.l, "Database not connected!")
 
+    # def enter_overnight_test(self):
```
archiepeterjarvis: If no longer required may be worth removing to clean up codebase.

PR #843, opened 2022-04-19
Incorporate fsck check into factory reset and factory console update

LettyYeti:
Also add broken and missing to list of errors.

### DIFF
--- a/src/asmcnc/apps/systemTools_app/screens/screen_factory_settings.py
+++ b/src/asmcnc/apps/systemTools_app/screens/screen_factory_settings.py
@@ -419,7 +419,7 @@ class FactorySettingsScreen(Screen):
     smartbench_model_path = '/home/pi/smartbench_model_name.txt'
     machine_serial_number_filepath  = "/home/pi/smartbench_serial_number.txt"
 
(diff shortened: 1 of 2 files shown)


src/asmcnc/apps/systemTools_app/screens/screen_factory_settings.py
```diff
@@ -685,12 +693,20 @@ def full_console_update(self):
...
+                self.console_update_button.text = 'Full Console Update (wifi)'
+                return False
+
             if self.set.get_sw_update_via_wifi():
                 self.set.fetch_platform_tags()
                 self.set.update_platform()
             else: 
                 message = "Could not get software update, please check connection."
                 popup_info.PopupWarning(self.systemtools_sm.sm, self.l, message)
+                self.console_update_button.text = 'Full Console Update (wifi)'
```
DennisKiselev, line 709: Does there not need to be a line like this in the if case directly above this else case?

LettyYeti: no, because it will do a full reboot as part of the update :)

PR #844, opened 2022-04-20
Maintenance bugfixes

DennisKiselev:
2 bugfixes:

- Clicking save in the touchplate offset tab in maintenance with an input outside of the permitted range now only shows 1 popup, rather than 2 as it did before
- Clicking save in the brush monitor tab in maintenance with an empty input field no longer crashes the console

### DIFF
--- a/src/asmcnc/apps/maintenance_app/widget_maintenance_brush_save.py
+++ b/src/asmcnc/apps/maintenance_app/widget_maintenance_brush_save.py
@@ -87,9 +87,23 @@ class BrushSaveWidget(Widget):
 
         # TIME FOR DATA VALIDATION
 
-        # Read from screen and convert hours into seconds
-        use = float(self.sm.get_screen('maintenance').brush_use_widget.brush_use.text)*3600
-        lifetime = float(self.sm.get_screen('maintenance').brush_life_widget.brush_life.text)*3600 
+        # Read from screen
+        use_str = self.sm.get_screen('maintenance').brush_use_widget.brush_use.text
+        lifetime_str = self.sm.get_screen('maintenance').brush_life_widget.brush_life.text
+
+        if use_str == '' or lifetime_str == '':
+            # throw popup, return without saving
+            warning_message = (
+                    self.l.get_str("There was a problem saving your settings.") + \
+                    "\n\n" + \
+                    self.l.get_str("Please check your settings and try again, or if the problem persists please contact the YetiTool support team.")
+                )
+            popup_info.PopupError(self.sm, self.l, warning_message)
+            return
+
+        #  Convert hours into seconds
+        use = float(use_str) * 3600
+        lifetime = float(lifetime_str) * 3600
 
         # Brush use
         if use >= 0 and use <= 999*3600: pass # all good, carry on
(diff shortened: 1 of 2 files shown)


src/asmcnc/apps/maintenance_app/widget_maintenance_brush_save.py
```diff
@@ -87,9 +87,23 @@ def save(self):
...
         # TIME FOR DATA VALIDATION
 
-        # Read from screen and convert hours into seconds
-        use = float(self.sm.get_screen('maintenance').brush_use_widget.brush_use.text)*3600
-        lifetime = float(self.sm.get_screen('maintenance').brush_life_widget.brush_life.text)*3600 
+        # Read from screen
+        use_str = self.sm.get_screen('maintenance').brush_use_widget.brush_use.text
+        lifetime_str = self.sm.get_screen('maintenance').brush_life_widget.brush_life.text
+
+        if use_str == '' or lifetime_str == '':
```
LettyYeti, line 94: rather than using an "if" here, use a try-except wrapped around the conversion. 

I managed to crash it by putting in a '-'.

src/asmcnc/apps/maintenance_app/widget_maintenance_z_misc_save.py
```diff
@@ -103,12 +103,14 @@ def get_info(self): # Localize me!
 
     def save(self):
```
LettyYeti, line 105: as per the conversation we had a little while back with arch, may be worth refactoring it to be more like: 

`if not save_touchplate_offset():`
`    warning popup `
`    return`
    
`if not save_z_head_maintenance():`
`    warning popup`
`    return`

`success popup`

DennisKiselev: I made the change but it actually seems more convoluted now, can you check if it seems right?

LettyYeti: ah I think I caused confusion with my pseudo-code - if there's just one opportunity for there to be a touchplate offset popup, either in the save_touchplate_offset function or in the save function, then it won't need the show_popup flag. 

So can either do: 

`if not save_touchplate_offset`
`warning popup with 1-2mm info`
`return False`

and 

`def save_touchplate_offset`
`if out of range:`
`return False`

OR

`if not save_touchplate_offset`
`return False`

and 

`def save_touchplate_offset`
`if out of range:`
`warning popup with 1-2mm info`
`return False`

does that clarify things?

DennisKiselev: But the flag is still surely needed, or else how will it know whether to show the general error popup or the out of range popup?

LettyYeti: ah I see what you mean - probably neatest to move the general error into the except of the save_touchplate_offset function I think

DennisKiselev: But what about the case where the save fails without an exception being thrown?

LettyYeti: you're right - I missed that one. 

Ok, this works well then, the only other alternative would be to [make it throw an exception](https://stackoverflow.com/questions/2052390/manually-raising-throwing-an-exception-in-python), but I'm not sure that's any tidier in this circumstance.

LettyYeti: ok in that case your first solution was deffo tidier, sorry! 😅

PR #879, opened 2022-06-20
Temp voltage ranges updates

LettyYeti:


src/asmcnc/comms/router_machine.py
```diff
@@ -1897,6 +1900,11 @@ def reset_tuning_flags(self):
         self.temp_toff = 2
         self.temp_sgt = 0
 
+    def motor_driver_temp_in_range(self, temp_to_assess):
+
+        if (self.lower_temp_limit <= temp_to_assess <= self.upper_temp_limit): return True
+        else: return False
+
```
archiepeterjarvis, line 1907: function could be simplified:

`def motor_driver_temp_in_range(self, temp_to_assess):
    return True if self.lower_temp_limit <= temp_to_assess <= self.upper_temp_limit else False`

src/asmcnc/production/z_head_qc_jig/z_head_qc_1.py
```diff
@@ -609,13 +609,13 @@ def temp_power_check(self, dt):
             fail_report.append("PCB Temperature: " + str(self.m.s.pcb_temp) + " degrees C")
             fail_report.append("Should be greater than 10 and less than 70 deg C.")
 
-        if 10 < self.m.s.motor_driver_temp < 60:
+        if 35 < self.m.s.motor_driver_temp < 100:
```
archiepeterjarvis, line 612: Maybe store this as variables somewhere? Not important but helps readability & maintenance

src/asmcnc/comms/router_machine.py
BenjiPestell: This guy's a genius

LettyYeti: good catch, corrected

archiepeterjarvis: Also as before you had it like 
if ___: return True 
else: return False 

you could do 

return True if ___ else return False 

which is a bit cleaner

PR #888, opened 2022-07-11
Alarm screen to handle stall guard/motor overload

LettyYeti:


src/asmcnc/core_UI/sequence_alarm/alarm_manager.py
```diff
@@ -210,8 +261,10 @@ def get_version_data(self):
 	def update_screens(self):
 
 		self.get_version_data()
-		if ((self.alarm_code).endswith('1') or (self.alarm_code).endswith('8')):
-			self.get_suspected_trigger()
+
+		if not self.sg_alarm: 
+			if ((self.alarm_code).endswith('1') or (self.alarm_code).endswith('8')):
```
archiepeterjarvis, line 266: Don't think extra brackets around the variable are required.

src/asmcnc/core_UI/sequence_alarm/alarm_manager.py
```diff
@@ -270,6 +322,8 @@ def setup_report(self):
 		except:
 			alarm_number = ""
 
+		if not self.sg_alarm: description = self.l.get_str(self.alarm_description)
+		else: description = (self.l.get_str("The N axis was overloaded during a move.")).replace("N", self.stall_axis)
```
archiepeterjarvis, line 327: `description = description = (self.l.get_str("The N axis was overloaded during a move.")).replace("N", self.stall_axis) if sg_alarm else self.l.get_str(self.alarm_description)`

Also when checking both cases of true/false with if/else for readability do if variable else not variable rather than vice versa

src/asmcnc/comms/serial_connection.py
```diff
@@ -876,6 +885,42 @@ def process_grbl_push(self, message):
...
+                    if 'Y' or 'y' in pins_info:
+
+                        # Depending on the firmware version (and the alarm type), 
+                        # Y pin means either Y max limit OR Y stall
+                        # and little y could be y home OR y limit
+                        if self.fw_version and int(self.fw_version.split('.')[0]) < 2:
+
+                            if 'y' in pins_info: self.limit_y = True
+                            else: self.limit_y = False
+
```
archiepeterjarvis, line 897: (think my comment disappeared?)

self.limit_y = 'y' in pins_info

src/asmcnc/core_UI/sequence_alarm/alarm_manager.py
LettyYeti: If most cases are going to be self.alarm_description, isn't is slightly more efficient to do:

description = self.l.get_str(self.alarm_description) if not self.sg_alarm else (self.l.get_str("The N axis was overloaded during a move.")).replace("N", self.stall_axis)

from a processing perspective? 

(I may be wrong) 

Also, is this the best place to put a line break: 

		description = self.l.get_str(self.alarm_description) if not self.sg_alarm \
		else (self.l.get_str("The N axis was overloaded during a move.")).replace("N", self.stall_axis)
		
?

archiepeterjarvis: I don't think so as it's already checking the value of sg_alarm so it's just a matter of 1/0 but either way probably very marginal. I just came from a readability perspective as you usually expect the NOT case to be in the else. 

And yes I think that line break works well.

PR #896, opened 2022-07-26
Job recovery

DennisKiselev:
Adds a new feature called job recovery, accessible from pro app after a job has been cancelled. Allows user to continue the job where they left off. Includes:

- Additions to job_data module to store information about cancelled job, including writing/reading text file to save info between power cycles, and gcode scrape function to recover position and modal gcodes at a certain point in a file
- Line selection screen where user can choose where to recover job from and move spindle to approximate location of current line
- Nudge screen that allows user to make minor adjustments in X and Y datums to account for position error

### DIFF
--- a/src/asmcnc/job/yetipilot/yetipilot.py
+++ b/src/asmcnc/job/yetipilot/yetipilot.py
@@ -183,7 +183,7 @@ class YetiPilot(object):
             command_delay = (self.override_command_delay if feed else self.spindle_override_command_delay) * i
 
             if feed:
-                if self.m.s.feed_override_percentage == 200:
+                if self.m.s.feed_override_percentage == 200 and adjustment > 0:
                     return adjustment_list[:i]
 
                 percentage_after_adjustments = adjustment + self.m.s.feed_override_percentage
--- a/src/main.py
+++ b/src/main.py
@@ -89,7 +89,7 @@ from asmcnc.skavaUI import screen_homing_decision # @UnresolvedImport
 Cmport = 'COM3'
 
 # Current version active/working on
-initial_version = 'v2.4.1'
+initial_version = 'v2.4.2'
 
 config_flag = False
         
--- a/v242.txt
+++ b/v242.txt
@@ -1,9 +1,9 @@
 .. _top:
 
 Precision Pro + upgrade capability, and YetiPilot improvements - with patch
-=============================================================================
+============================================================================
 
-This is a patch for the YetiPilot release, v2.4.0: we’ve added units (mm) back into cutter diameter and step down fields for all profiles.
+This is a patch for the YetiPilot release, v2.4.0: we’ve fixed an issue where the YetiPilot feed override settings would sometimes stick. 
 
 v2.4.0 gives existing owners of SmartBench V1.3 the ability to upgrade their machines to Precision Pro + with an upgrade kit.
 


src/asmcnc/job/job_data.py
```diff
@@ -389,3 +400,208 @@ def post_job_data_update_post_send(self):
...
+                    recovery_gcode.append("G0 X" + x + " Y" + y)
+                    recovery_gcode.append("G0 Z" + z)
+            else:
+                recovery_gcode.append("G0 X" + x + " Y" + y)
+                recovery_gcode.append("G0 Z" + z)
+                if feedrate_line:
+                    recovery_gcode.append("G1 F" + feedrate)
+
+            # Coordinate System Select
+            coord_system_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("G5[4-9]", s)), None)
```
LettyYeti, line 515: Does this need to be set before the X Y Z positions?

src/asmcnc/job/job_data.py
```diff
@@ -389,3 +400,208 @@ def post_job_data_update_post_send(self):
...
+                    return False, 'This job contains the G91 g-code, so it cannot be recovered.'
+                recovery_gcode.append("G90")
+
+            # Arc IJK distance mode
+            arc_mode_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("G9[0,1]\.1", s)), None)
+            if arc_mode_line:
+                recovery_gcode.append(re.search("G9[0,1]\.1", arc_mode_line).group(0))
+
+            # Feed rate mode
+            feedrate_mode_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("G9[3-5]", s)), None)
```
LettyYeti, line 538: does this need setting before feed rate?

src/asmcnc/job/job_data.py
```diff
@@ -389,3 +400,208 @@ def post_job_data_update_post_send(self):
...
+            if feedrate_mode_line:
+                recovery_gcode.append(re.search("G9[3-5]", feedrate_mode_line).group(0))
+
+            # Units
+            unit_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("G2[0,1]", s)), None)
```
LettyYeti, line 543: before coords?

src/asmcnc/skavaUI/screen_go.py
```diff
@@ -599,20 +599,43 @@ def _start_running_job(self):
...
+            if ((str(self.jd.job_gcode).count("M3") > str(self.jd.job_gcode).count("M30")) or (
+                    str(self.jd.job_gcode).count("M03") > 0)) and self.m.stylus_router_choice != 'stylus':
+                modified_job_gcode.append("AE")  # turns vacuum on
+                modified_job_gcode.append("G4 P2")  # sends pause command
+                modified_job_gcode.extend(self.jd.job_gcode)
+                modified_job_gcode.append("G4 P2")  # sends pause command, 2 seconds
+                modified_job_gcode.append("AF")  # turns vac off
+                self.jd.job_recovery_offset += 2 # number of lines added to start of job
+            else:
+                modified_job_gcode.extend(self.jd.job_gcode)
```
LettyYeti, line 639: both of these methods look the same, so suggest can make this more readable:

```
if job recovery conditions:
    modified_job_gcode = self.prep_gcode(self.jd.job_recovery_gcode)
else: 
    self.jd.job_recovery_offset = 0
    modified_job_gcode = self.prep_gcode(self.jd.job_gcode)

def prep_gcode(job_to_modify):
    if self.lift_z_on_job_pause and self.m.fw_can_operate_zUp_on_pause(): etc....
```

src/asmcnc/skavaUI/screen_home.py
```diff
@@ -207,9 +210,28 @@
                                             size: self.parent.width, self.parent.height
                                             allow_stretch: True
 
+                                Button:
```
LettyYeti, line 213: This needs to be commented out on merge (or disabled & made invisible)

src/asmcnc/skavaUI/widget_z_move_recovery.py
```diff
@@ -0,0 +1,205 @@
...
+    
+    def jogModeCycled(self):
+
+        self.jog_mode_button_press_counter += 1
+        if self.jog_mode_button_press_counter % 6 == 0: 
+            self.jogMode = 'free'
+            self.jogModeButtonImage.source = './asmcnc/skavaUI/img/jog_mode_infinity.png'
+        if self.jog_mode_button_press_counter % 6 == 1: 
+            self.jogMode = 'job'
+            self.jogModeButtonImage.source = './asmcnc/skavaUI/img/jog_mode_box.png'
```
LettyYeti, line 151: don't think it needs the job option

src/asmcnc/skavaUI/screen_job_recovery.py
```diff
@@ -0,0 +1,419 @@
...
+        if self.jd.job_recovery_selected_line == -1:
+            self.line_input.text = ""
+            # Take away 1 so that user can't move to the position that was the target during cancellation
+            self.initial_line_index = self.jd.job_recovery_cancel_line - 1
+            self.selected_line_index = self.initial_line_index
+            self.display_list = ["" for _ in range (6)] + [str(i) + ": " + self.jd.job_gcode[i] for i in range(self.initial_line_index + 1)] + ["" for _ in range (6)]
+
+            self.stopped_on_label.text = "Job failed during line " + str(self.initial_line_index)
+            self.display_list[self.selected_line_index + 6] = "[color=FF0000]" + self.display_list[self.selected_line_index + 6] + "[/color]"
+            self.update_display()
```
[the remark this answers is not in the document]
DennisKiselev: When the edge of the list is reached it needs to show 6 empty lines above the first one and vice versa at the bottom, or else the label wouldn't be centered properly and look a bit weird while you scroll

src/asmcnc/skavaUI/screen_job_recovery.py
```diff
@@ -0,0 +1,419 @@
...
+
+    def scroll_down(self):
+        if self.selected_line_index < self.initial_line_index:
+            self.selected_line_index += 1
+            self.update_display()
+
+    def jump_to_line(self, instance, value):
+        if value:
+            if value == "-":
```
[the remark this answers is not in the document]
DennisKiselev: Yep makes no difference, updated anyway for robustness

src/asmcnc/skavaUI/screen_job_recovery.py
```diff
@@ -0,0 +1,419 @@
...
+        if y_line:
+            self.pos_y = float(re.split('(X|Y|Z|F|S|I|J|K|G)', y_line)[re.split('(X|Y|Z|F|S|I|J|K|G)', y_line).index('Y') + 1])
+        else:
+            self.pos_y = 0.0
+        z_line = next((s for s in reversed(self.jd.job_gcode[:self.selected_line_index + 1]) if 'Z' in s), None)
+        if z_line:
+            self.pos_z = float(re.split('(X|Y|Z|F|S|I|J|K|G)', z_line)[re.split('(X|Y|Z|F|S|I|J|K|G)', z_line).index('Z') + 1])
+        else:
+            self.pos_z = 0.0
+
```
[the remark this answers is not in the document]
DennisKiselev: Will keep in mind - did it this way because it saves a bit of code but I understand readability and testability being more important

src/asmcnc/skavaUI/screen_job_recovery.py
```diff
@@ -0,0 +1,419 @@
...
+        self.m.s.write_command('G90 G0 X%s Y%s' % (self.pos_x, self.pos_y))
+
+    def back_to_home(self):
+        self.jd.reset_recovery()
+        self.sm.current = 'home'
+
+    def next_screen(self):
+        if self.m.state().startswith("Idle"):
+            self.wait_popup = popup_info.PopupWait(self.sm, self.l)
+            self.jd.job_recovery_selected_line = self.selected_line_index + 1
```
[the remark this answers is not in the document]
DennisKiselev: Was asked to make it so that the selected line is included in the scrape and then have it continue on from the next line. Bit more clear when the position and speed info at the bottom reflects the selected line rather than the one before it I guess

src/asmcnc/skavaUI/widget_z_move_recovery.py
DennisKiselev: Whoops

src/asmcnc/job/job_data.py
DennisKiselev: Moved modal gcode scraping to before speed/coord scraping, should fix these next 3 issues

src/asmcnc/skavaUI/screen_go.py
```diff
@@ -597,49 +597,38 @@ def _start_running_job(self):
...
+                modified_job_gcode.append("M56 P0")  # append cleaned up gcode to object
+
+            return modified_job_gcode
 
-        # Spindle command??
-        if self.lift_z_on_job_pause and self.m.fw_can_operate_zUp_on_pause():  # extra 'and' as precaution
-            modified_job_gcode.append("M56 P0")  # append cleaned up gcode to object
+        # Check if job recovery is being performed
+        if self.jd.job_recovery_filepath == self.jd.filename and self.jd.job_recovery_selected_line != -1:
+            modified_job_gcode = prep_gcode(self.jd.job_recovery_gcode)
```
LettyYeti, line 628: I would have modified_job_gcode in the inner prep_gcode function renamed, as I think both versions of the variable might exist within the same scope as prep_gcode is a nested function. I think it would be ok (as it ultimately ends up being the same list anyway), but might still be a good idea to rename :).

LettyYeti (commented):
See comment on go screen :)

src/asmcnc/skavaUI/screen_job_recovery.py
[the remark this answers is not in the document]
LettyYeti: ha glad I asked! Would not have been able to guess 😂

src/asmcnc/skavaUI/screen_job_recovery.py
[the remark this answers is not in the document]
LettyYeti: does that mean that the line you select to go from isn't actually the line that's executed? It's the next line that's executed?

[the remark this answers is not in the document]
DennisKiselev: Yep exactly

src/asmcnc/job/job_data.py
```diff
@@ -389,3 +400,209 @@ def post_job_data_update_post_send(self):
...
+    def generate_recovery_gcode(self):
+        try:
+            recovery_gcode = []
+
+            # Recover modal gcodes
+
+            # Coordinate System Select
+            coord_system_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("G5[4-9]", s)), None)
+            if coord_system_line:
+                recovery_gcode.append(re.search("G5[4-9](\.[1-3])?", coord_system_line).group(0))
```
LettyYeti, line 467: If no G54, should it give user a warning? 

In this scenario, it would return True and a message

@Neex101

src/asmcnc/job/job_data.py
```diff
@@ -389,3 +400,209 @@ def post_job_data_update_post_send(self):
...
+
+            # Plane selection
+            plane_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("G1[7-9]", s)), None)
+            if plane_line:
+                recovery_gcode.append(re.search("G1[7-9]", plane_line).group(0))
+
+            # Distance mode
```
LettyYeti, line 474: Make comments # Absolute or incremental distance mode

src/asmcnc/job/job_data.py
```diff
@@ -389,3 +400,209 @@ def post_job_data_update_post_send(self):
...
+            distance_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("G9[0,1]([A-Z]|\s|$)", s)), None)
+            if distance_line:
+                # Recovery not allowed if G91 is used
+                if re.search("G91([A-Z]|\s|$)", distance_line):
+                    return False, 'This job contains the G91 g-code, so it cannot be recovered.'
```
LettyYeti, line 479: Change error message to: 

"The last positioning declaration was incremental (G91), and therefore this job cannot be recovered."

src/asmcnc/job/job_data.py
```diff
@@ -389,3 +400,209 @@ def post_job_data_update_post_send(self):
...
+                recovery_gcode.append("G90")
+
+            # Arc IJK distance mode
+            arc_mode_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("G9[0,1]\.1", s)), None)
```
LettyYeti, line 483: @Neex101 - we need to confirm what to do on these. G91.1 probably still needs to be binned (as with G91), but maybe we can handle G91.1 ? That may be conditional.

src/asmcnc/job/job_data.py
```diff
@@ -389,3 +400,209 @@ def post_job_data_update_post_send(self):
...
+            if arc_mode_line:
+                recovery_gcode.append(re.search("G9[0,1]\.1", arc_mode_line).group(0))
+
+            # Feed rate mode
+            feedrate_mode_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("G9[3-5]", s)), None)
+            if feedrate_mode_line:
+                recovery_gcode.append(re.search("G9[3-5]", feedrate_mode_line).group(0))
```
LettyYeti, line 490: We're not sure how to handle G93 or G95, so return False for these (and error message... @Neex101 any preferences on what that should be?). G94 only for time being :)

src/asmcnc/job/job_data.py
```diff
@@ -389,3 +400,209 @@ def post_job_data_update_post_send(self):
...
+            if next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if 'G40' in s), None):
+                recovery_gcode.append('G40')
+
+            # Tool length offset
+            tool_length_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("G4(3\.1|9)", s)), None)
+            if tool_length_line:
+                recovery_gcode.append(re.search("G4(3\.1|9)", tool_length_line).group(0))
+
+            # Program mode
+            program_line = next((s for s in reversed(self.job_gcode[:self.job_recovery_selected_line]) if re.search("M0?[0-2](\D|$)|M30", s)), None)
```
LettyYeti, line 507: @BenjiPestell @benjipestell64 Can you do a test with X rack cutter file with pause to ensure this works in the way we want? Concerned that it will pause at job beginning, even if pause was cleared prior to cancellation.

BenjiPestell: I had Dennis put in a fix for this, works fine now

src/asmcnc/job/job_data.py
LettyYeti: @DennisKiselev I think the correct logic would be to throw an error if G91.1 is the last distance mode used, as it's anotther kind of incremental mode for arcs. 

> G91.1 – Incremental distance mode for I and K offsets. G91.1 returns I and K to their default behavior.

For this merge, it might be best to throw an error if trying to recover a job with G90.1 or G2/G3 as last modes as well, as these are arcs and G2/3 and G90.1 need handling together, but I think we could put in the logic for handling that later on if we wanted.

> G90.1 – Absolute distance mode for I and K offsets. When G90.1 is in effect, I and K both must be specified with G02/G03 for the XZ plane or it is an error.

@Neex101 any thoughts on the above?

src/asmcnc/job/job_data.py
LettyYeti: maybe a warning message

src/asmcnc/job/job_data.py
LettyYeti: throw error

src/asmcnc/job/job_data.py
LettyYeti: No warning required - have found various files in the gcode stash that don't explicitly state G54, so we can assume it's fine :)

src/asmcnc/job/job_data.py
LettyYeti: Error message along lines of: 

"Job recovery only supports feed rate mode G94. This job contains [G93/G95], and therefore cannot be recovered."

src/asmcnc/job/job_data.py
LettyYeti: "Job recovery does not currently support arc distance modes. This job contains [G90.1/G91.1], and therefore cannot be recovered."

LettyYeti:
Realised I wasn't clear re the error message - can you set them up so that they return the specific G-code that the file contains? For example, rather than "[G90.1/G91.1]", the error message says "This job contains G91.1" OR "This job contains G90.1"

PR #899, opened 2022-07-31
Set Z current to 25 when v1.3 Z Heads are plugged into QC jig

LettyYeti:
- Also ensured that registers and FW/HW versions are cleared/reset when serial is disconnected and reconnected (as Z Head will change) 
- Ensure that "Z Head connecting" screen (which also sets the current) comes up after reconnection called from warranty screen as well as after full QC

### DIFF
--- a/src/asmcnc/comms/router_machine.py
+++ b/src/asmcnc/comms/router_machine.py
@@ -170,12 +170,16 @@ class RouterMachine(object):
         if self.s.is_connected():
             self.s.s.close()
         self.clear_motor_registers()
+        self.s.fw_version = ''
+        self.s.hw_version = ''
         self.s.establish_connection(self.win_serial_port)
 
     def close_serial_connection(self, dt):
         if self.s.is_connected():
             self.s.s.close()
         self.clear_motor_registers()
+        self.s.fw_version = ''
+        self.s.hw_version = ''
 
 # PERSISTENT MACHINE VALUES
     def check_presence_of_sb_values_files(self):
--- a/src/asmcnc/comms/router_machine.py
+++ b/src/asmcnc/comms/router_machine.py
@@ -167,11 +167,7 @@ class RouterMachine(object):
 
     # CREATE/DESTROY SERIAL CONNECTION (for cycle app)
     def reconnect_serial_connection(self):
-        if self.s.is_connected():
-            self.s.s.close()
-        self.clear_motor_registers()
-        self.s.fw_version = ''
-        self.s.hw_version = ''
+        self.close_serial_connection(0)
         self.s.establish_connection(self.win_serial_port)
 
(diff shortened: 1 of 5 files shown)


src/asmcnc/comms/router_machine.py
```diff
@@ -169,11 +169,17 @@ def __init__(self, win_serial_port, screen_manager, settings_manager, localizati
     def reconnect_serial_connection(self):
         if self.s.is_connected():
```
DennisKiselev, line 170: First 5 lines of this function are the same as close_serial_connection, could just call close_serial_connection instead to avoid duplicate code?

src/asmcnc/production/z_head_qc_jig/z_head_qc_connecting.py
```diff
@@ -52,45 +52,58 @@ def __init__(self, **kwargs):
...
 
 
     def get_and_set_current(self):
 
-    	if not self.m.s.fw_version:
+        if not self.m.s.fw_version:
+            Clock.schedule_once(lambda dt: self.get_and_set_current(), 0.5)
+            return
+
+        # If current is already set to 22, carry onto QC home
```
DennisKiselev, line 84: This comment seems out of place

src/asmcnc/comms/router_machine.py
```diff
@@ -2839,11 +2847,34 @@ def set_motor_current(self, axis, current):
                 time.sleep(0.5)
 
             self.send_command_to_motor("STORE TMC PARAMS IN EEPROM", command = STORE_TMC_PARAMS)
-            time.sleep(0.5)
+            time.sleep(1)
             self.tmc_handshake()
-            time.sleep(0.5)
+            time.sleep(1)
             return True
```
archiepeterjarvis, line 2853: Is time.sleep() safe? I was always told to use a clock, good conversation either way.

LettyYeti: I've used it for these functions because I want a blocking delay after commands have been sent (so serial comms thread is not blocked), and don't mind in this instance if it locks up the UI (as we're just doing it on a plain screen at start of Z Head QC). If we were to use it in something customer facing, we might want something that uses clocks (so that UI stays responsive). 

That said, we also don't want any other commands sent while TMC PARAMS are being received, so the block is useful for that reason too. 

So yes, clocks are preferable most of the time, but in this instance I've still chosen to use sleep.

src/asmcnc/comms/router_machine.py
LettyYeti: 👍

src/asmcnc/comms/router_machine.py
archiepeterjarvis: yep, I figured that would be the case. Thanks

PR #900, opened 2022-08-01
Auto log

archiepeterjarvis:
Functionality not pulled into anything yet, but has been fully tested.


src/asmcnc/comms/logging/log_exporter.py
```diff
@@ -0,0 +1,53 @@
...
+    if not os.path.exists(export_logs_folder) or not os.path.isdir(export_logs_folder):
+        os.mkdir(export_logs_folder)
+
+
+def generate_logs():
+    create_log_folder()
+
+    str_current_time = datetime.now().strftime("%m-%d-%Y-%H-%M-%S")
+
+    log_name = "easycut-exported-logs-" + str_current_time + ".txt"
```
LettyYeti, line 21: Is there a reason not to use the smartbench_logs.txt path (/home/pi/smartbench_logs.txt) that is generated when users download logs with onto USB? Just thinking it would be good to keep all logs in the same place (also for my sanity when I look for them), and it means that any big log files won't get duplicated.

LettyYeti: time looks useful to have in file as well though

src/asmcnc/comms/logging/log_exporter.py
```diff
@@ -0,0 +1,53 @@
...
+    full_path = export_logs_folder + '/' + log_name
+
+    command = "journalctl > " + full_path
+
+    os.system(command)
+
+    return log_name
+
+
+def trim_logs(log_file_path, x_lines):
```
LettyYeti, line 31: purpose of this?

archiepeterjarvis: me and Ed discussed, may be useful for future application to be able to trim down. e.g. if we used this system for automatic customer support logs, it would be easier for less familiar people like Elliot if there wasn't thousands of lines to go through

src/asmcnc/comms/logging/log_exporter.py
archiepeterjarvis: no particular reason, but time is helpful. Will add a serial too if possible

LettyYeti: ok, can you rename it smartbench-exported-logs- (time) - (serial) .txt ? 

then we can make the USB export match the same location later on. Customers don't know that the SW is called easycut, and it can be confusing.

archiepeterjarvis:
Tested on rig by going through calibration procedures and checking whether log file is transferred.

archiepeterjarvis:
Tested on rigs & machine going through calibration procedure, making sure file is trimmed to correct length.

src/asmcnc/comms/logging/log_exporter.py
```diff
@@ -0,0 +1,113 @@
+from datetime import datetime
+import os
+import paramiko
```
LettyYeti, line 3: put paramiko import into try-except. we install it with cryptography on the pis, but will be good for ensuring dev systems can run without freaking out :)

src/asmcnc/comms/logging/log_exporter.py
```diff
@@ -0,0 +1,113 @@
...
+import sys
+
+WORKING_DIR = 'A:\\Logs\\'
```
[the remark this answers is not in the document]
archiepeterjarvis: yep

src/asmcnc/production/z_head_qc_jig/z_head_qc_connecting.py
```diff
@@ -55,6 +55,12 @@ def __init__(self, **kwargs):
         self.m = kwargs['m']
         self.usb = kwargs['usb']
         self.connecting_label.text = "Connecting to Z Head..."
+        self.z_current = 25
+        self.x_current = 26
+
+        # IMPORTANT: uncomment below to run z head qc on riggy
+        # self.z_current = 5
+        # self.x_current = 5
```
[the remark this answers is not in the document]
LettyYeti: @archiepeterjarvis this is interesting, as this bit isn't in master anymore. I'm not sure why it's persisting?

PR #916, opened 2022-08-21
Upgrade tuning and calibration functions to work with dual X driver PCBs

LettyYeti:


src/asmcnc/comms/router_machine.py
```diff
@@ -2157,15 +2163,19 @@ def do_tuning(self, X, Y, Z):
...
-
         try: 
 
             tuning_array, current_temp = self.sweep_toff_and_sgt_and_motor_driver_temp(X = X, Y = Y, Z = Z)
             log("Sweep finished")
 
             if X: 
                 X_target_SG = self.get_target_SG_from_current_temperature('X', current_temp)
-                self.x_toff_tuned, self.x_sgt_tuned = self.find_best_combo_per_motor_or_axis(tuning_array, X_target_SG, 1)
+                if self.s.sg_x1_motor and self.s.sg_x2_motor:
```
DennisKiselev, line 2173: Are these guaranteed to not have a value of 0? If not then should be changed to `if (self.s.sg_x1_motor != None) and (self.s.sg_x2_motor != None):`

LettyYeti: good catch, you're absolutely right! will fix now :)

LettyYeti: @DennisKiselev have updated, all good?

DennisKiselev: Yep all good!

PR #917, opened 2022-08-22
Systematically went through to ensure SW will work with both 4 and 5 drivers :) 

LettyYeti:


src/asmcnc/apps/systemTools_app/screens/calibration/screen_current_adjustment.py
```diff
@@ -199,11 +229,22 @@ def home(self):
         self.m.request_homing_procedure('current_adjustment','current_adjustment')
 
     def measure(self):
+        
         if self.m.s.sg_x_motor_axis != -999:
             self.x_vals.append(self.m.s.sg_x_motor_axis)
             self.rt_x_sg.text = str(self.m.s.sg_x_motor_axis)
             self.peak_x_sg.text = str(max(self.x_vals))
 
+        if self.m.s.sg_x1_motor != -999 or None:
```
DennisKiselev, line 238: This reads None as a condition, should be `if self.m.s.sg_x1_motor != -999 or self.m.s.sg_x1_motor != None:` or `if self.m.s.sg_x1_motor in [-999, None]:`

LettyYeti: fixed :)

PR #918, opened 2022-08-22
Job recovery redo

DennisKiselev:
- Completing a job now stores info about the job, so that the job can be quickly repeated even if recovery isn't needed
- Jobs are now loaded through job recovery, so the correct job does not need to be loaded to perform job recovery
- Z datum reminder reset in job recovery (through loading screen)
- Shapecutter job info no longer gets stored in recovery file

### DIFF
--- a/src/asmcnc/skavaUI/screen_home.py
+++ b/src/asmcnc/skavaUI/screen_home.py
@@ -341,20 +341,6 @@ class HomeScreen(Screen):
             except:
                 log('Unable to preview file')
 
-            # Check if job recovery is available
-            if self.jd.job_recovery_cancel_line > 0 and self.jd.job_recovery_filepath == self.jd.filename:
-                self.job_recovery_button.disabled = False
-                self.job_recovery_button_image.source = "./asmcnc/skavaUI/img/recover_job.png"
-
-                # Line -1 being selected represents no selected line
-                if self.jd.job_recovery_selected_line == -1:
-                    self.file_data_label.text += "\n[color=FF0000]Restart from beginning[/color]"
-                else:
-                    self.file_data_label.text += "\n[color=FF0000]From line " + str(self.jd.job_recovery_selected_line) + "[/color]"
-            else:
-                self.job_recovery_button.disabled = True
-                self.job_recovery_button_image.source = "./asmcnc/skavaUI/img/recover_job_disabled.png"
-
     def on_pre_enter(self):
 
         if self.jd.job_gcode == []:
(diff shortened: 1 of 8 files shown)


src/asmcnc/skavaUI/screen_home.py
```diff
@@ -376,10 +349,39 @@ def on_pre_enter(self):
...
+            # Cancel on line -1 represents last job completing successfully
+            if self.jd.job_recovery_cancel_line == -1:
+                self.job_recovery_button_image.source = "./asmcnc/skavaUI/img/recover_job_disabled.png"
+
+            else:
+                self.job_recovery_button_image.source = "./asmcnc/skavaUI/img/recover_job.png"
+
+            # Line -1 being selected represents no selected line
+            if self.jd.job_recovery_selected_line == -1:
+                if self.jd.job_recovery_from_beginning:
```
LettyYeti, line 379: what does this flag do that the == -1 check doesn't do?

src/asmcnc/skavaUI/screen_job_recovery.py
```diff
@@ -398,7 +398,7 @@ def go_xy(self):
         self.m.s.write_command('G90 G0 X%s Y%s' % (self.pos_x, self.pos_y))
 
     def back_to_home(self):
-        self.jd.reset_recovery()
+        self.jd.reset_values()
         self.sm.current = 'home'
```
LettyYeti, line 402: does reset_recovery function still get used anywhere? does it need clearing from job data if not?

LettyYeti (commented):
See comments :)

src/asmcnc/skavaUI/screen_job_recovery.py
DennisKiselev: Yep, it's used when writing to the recovery file

src/asmcnc/skavaUI/screen_home.py
DennisKiselev: -1 represents no selected line but not necessarily restarting from the beginning (e.g. no file is loaded, so no line is selected but not restarting from beginning) so extra flag needed to differentiate the cases

src/asmcnc/skavaUI/screen_home.py
```diff
@@ -376,10 +349,39 @@ def on_pre_enter(self):
                 self.gcode_preview_widget.draw_file_in_xy_plane([])
                 self.gcode_preview_widget.get_non_modal_gcode([])
             except:
-                print 'No G-code loaded.'
+                print('No G-code loaded.')
 
             self.gcode_summary_widget.hide_summary()
 
+        else:
```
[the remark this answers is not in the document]
LettyYeti: In that case, can you just move the bits that are necessary for job recovery/filename display? and leave the gcode summary and preview where they were?

src/asmcnc/skavaUI/screen_home.py
```diff
@@ -363,23 +362,29 @@ def on_pre_enter(self):
...
-            self.job_box.range_z[1] = 0
-            
-            # Hack to clear any previous job files
-            try:            
-                self.gcode_preview_widget.draw_file_in_xy_plane([])
-                self.gcode_preview_widget.get_non_modal_gcode([])
-            except:
-                print 'No G-code loaded.'
 
-            self.gcode_summary_widget.hide_summary()
```
LettyYeti, line 381: This stuff still looks like it's not where it was before, and I don't think job recovery is dependent on it? 

Might be worth using master as a reference for how any previews should be :)

PR #926, opened 2022-09-05
Zh qc fw

DennisKiselev:
Only allow FW suitable to current HW to be flashed

### DIFF
--- a/src/asmcnc/production/z_head_qc_jig/z_head_qc_home.py
+++ b/src/asmcnc/production/z_head_qc_jig/z_head_qc_home.py
@@ -91,12 +91,11 @@ class ZHeadQCHome(Screen):
             return "/media/usb/GRBL*5.hex"
 
         elif int(self.m.s.hw_version) >= 20:
-            try:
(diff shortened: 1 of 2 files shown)


src/asmcnc/production/z_head_qc_jig/z_head_qc_home.py
```diff
@@ -86,9 +86,22 @@ def __init__(self, **kwargs):
...
+            return "/media/usb/GRBL*5.hex"
+
+        elif int(self.m.s.hw_version) >= 20:
+            if glob.glob("/media/usb/GRBL*4.hex"):
+                return "/media/usb/GRBL*4.hex"
+
+            # Allow for older FW versions, starting in 2 but not ending in 4
+            return "/media/usb/GRBL23*.hex"
+
+        return "/media/usb/GRBL*.hex"
```
LettyYeti, line 100: If this is GRBL1*, this will mean that multiple files can exist on same USB I think, so may as well :)

PR #931, opened 2022-09-07
Z head mechanics jig

DennisKiselev:
New app that performs a test that records load values while Z axis is moving up and down, and plots them in a graph against their Z position.


src/asmcnc/production/z_head_mechanics_jig/z_head_mechanics.py
```diff
@@ -0,0 +1,369 @@
+from kivy.uix.screenmanager import Screen
+from kivy.lang import Builder
+from kivy.clock import Clock
+
+from asmcnc.comms.yeti_grbl_protocol.c_defines import *
+from asmcnc.skavaUI import popup_info
+
+import matplotlib.pyplot as plt
```
LettyYeti, line 8: matplotlib MUST be included in ansible package install list, and this must be tested (you will need to uninstall and clear any cached install files in order to test)

src/asmcnc/production/z_head_mechanics_jig/z_head_mechanics.py
```diff
@@ -0,0 +1,369 @@
...
+        plt.tight_layout()
+        plt.grid()
+        plt.savefig('./asmcnc/production/z_head_mechanics_jig/z_head_mechanics_jig_graph.png')
+        plt.close()
+        self.load_graph.source = './asmcnc/production/z_head_mechanics_jig/z_head_mechanics_jig_graph.png'
+        self.load_graph.reload()
+        self.load_graph.opacity = 1
+
+
+    def stop(self):
```
LettyYeti, line 314: This won't prevent any scheduled events from continuing to run (i.e. anything that's on a Clock). Possible that movements stop, GRBL gets reset, and then events will carry on & start new movements.

PR #946, opened 2022-09-21
Idle current adjustment

DennisKiselev:
Idle current is now set along with active current in current adjustment screen.


DennisKiselev:
Tested on rig, made sure that both idle and active current commands are sent on each button press, and that after storing and rebooting, registers show the correct currents

PR #947, opened 2022-09-23
Up down jig current input

DennisKiselev:
Added text inputs to manual move screen to allow user to customise current in both phases of the test. The current is also displayed on the main screen whenever it is changed


DennisKiselev:
Tested using rig, made sure that current is correctly displayed on main screen when current hasn't been changed, and made sure that it also changes to the correct values after updating them in manual move screen. Having issues with Z motor though, can't get it to work without the SET_TOFF command after every startup. But this is unrelated to anything I did on this branch so it should work fine on a proper rig?

PR #953, opened 2022-09-26
Fix matplotlib

LettyYeti:


LettyYeti:
Tested on rig before and after matplotlib install :)

PR #956, opened 2022-09-29
Coefficient fix

LettyYeti:


LettyYeti:
Tested on rig + ZH, did not try sending to DB (values were out of range), but could see in terminal information that cal coeffs were printed correctly after reset

PR #959, opened 2022-10-05
Second cal check

LettyYeti:


LettyYeti:
Testing:

Tested on dev machine from end of stall jig, and confirmed that data was in database after it was sent

PR #961, opened 2022-10-10
Stall alarm reminder

DennisKiselev:
Stall alarm screen now has a reminder about raising the Z axis


DennisKiselev:
Unit tested using alarm_screen_tests.py, made sure text fully fits on every language

PR #962, opened 2022-10-11
Show popup if update button is pressed with $54 set wrong

DennisKiselev:


DennisKiselev:
Tested on rig, made sure that popup appears when $54=1 and that update works if $54=0

PR #963, opened 2022-10-11
Swapped buttons

DennisKiselev:
Swapped reboot and support buttons on system tools, so that accidentally double clicking the system tools button puts you on the support screen rather than rebooting


DennisKiselev:
Tested on windows, made sure that buttons are swapped and function as they should

PR #965, opened 2022-10-14
Final test yellow screens

LettyYeti:
We were seeing intermittent gcode errors in overnight test, between recalibration stage and full calibrated run stage. 

This was caused by setting the datum prior to file streaming, which also had LED commands sent on a clock. These LED commands then interfered with the file stream, causing a gcode error 😳


LettyYeti:
Tested in mini run dev mode on dev machine

PR #968, opened 2022-10-20
Digital spindle overload

DennisKiselev:
Added functionality to serial comms to update internal overload value using killtime when SC2 spindle is in use, to be displayed on the go screen


DennisKiselev:
Virtually tested using a unit test, physically tested by Benji

LettyYeti:
SC1 overload tested with riggy

PR #972, opened 2022-10-25
block update if machine state is not Idle

LettyYeti:


LettyYeti:
Tested on ZH QC jig with a single driver v3 PCB

PR #973, opened 2022-10-28
Set different hostnames for pi4 and pi3

LettyYeti:
If Pi4, serial has leading 100000*..., and Pi 3 has leading 000000*. For pi 4, prepend a 4 onto last 8 digit of SN to differentiate it from pi 3 SNs. 

Consequences: pi4 hostnames will be 9 characters instead of 8.


LettyYeti:
Tested on Pi 3 
Tested on Pi 4

LettyYeti:
On the pi 3 also double checked that the new script had definitely copied across and ran (output was the same)

On pi 4, started out with pi@raspberry so saw change on restart (but also checked script had copied)

PR #986, opened 2022-11-14
Always write to recovery file if enough time has passed

DennisKiselev:
When cancelling a job, recovery info is written if more than 30 seconds have passed since the start of the job, even if the cancel line is less than 0. This means that files that calculate the cancel line incorrectly (i.e. file containing G2/G3) will still have recovery info saved.

Tested on rig

### DIFF
--- a/src/asmcnc/comms/serial_connection.py
+++ b/src/asmcnc/comms/serial_connection.py
@@ -619,16 +619,16 @@ class SerialConnection(object):
             Clock.schedule_once(lambda dt: self.m.zUp(), 0.5)
             Clock.schedule_once(lambda dt: self.m.vac_off(), 1)
 
+            # Update time for maintenance reminders
+            time.sleep(0.4)
+            time_taken_seconds = self.update_machine_runtime()
+
             # Write recovery info
             # g_count represents the number of OKs issued by grbl which are sent when a line enters the line buffer
             # The line buffer has a capacity of 35 lines
             # So the currently executing command is the one 35 lines before the last one received by the buffer
             if not self.jd.job_recovery_skip_recovery:
-                self.jd.write_to_recovery_file_after_cancel(self.g_count - 35, self.sm.get_screen('go').time_taken_seconds)
-
-            # Update time for maintenance reminders
-            time.sleep(0.4)
-            self.update_machine_runtime()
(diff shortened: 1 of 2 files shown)


src/asmcnc/job/job_data.py
```diff
@@ -422,14 +422,12 @@ def read_from_recovery_file(self):
...
-    def write_to_recovery_file_after_cancel(self, cancel_line):
+    def write_to_recovery_file_after_cancel(self, cancel_line, cancel_time):
         try:
             # Account for number of lines added in by the software when running file
             cancel_line -= self.job_recovery_offset
 
-            # If job was cancelled before it started, no need to store recovery info
-            if cancel_line >= 0:
-
+            def write_to_file(cancel_line):
```
LettyYeti, line 430: I think move the function definition out of the try-except clause (so above the try:), as any failure at run time will still be caught, and it makes everything a little more readable.

src/asmcnc/comms/serial_connection.py
```diff
@@ -624,7 +624,7 @@ def cancel_stream(self):
             # The line buffer has a capacity of 35 lines
             # So the currently executing command is the one 35 lines before the last one received by the buffer
             if not self.jd.job_recovery_skip_recovery:
-                self.jd.write_to_recovery_file_after_cancel(self.g_count - 35)
+                self.jd.write_to_recovery_file_after_cancel(self.g_count - 35, self.sm.get_screen('go').time_taken_seconds)
```
LettyYeti, line 628: I would like, in future, to try and decouple the screens from the backend (I'm aware that at the moment it's a right mess and screens are pulled into the serial comms all over the place)

I think it might be better if the update_machine_runtime() function is called first, you can modify it to return time_taken_seconds at the end of the function, and then this value can be passed to the write_to_recovery function instead.

src/asmcnc/job/job_data.py
LettyYeti, line 430: I think move the function definition out of the try-except clause (so above the try:), as any failure at run time will still be caught, and it makes everything a little more readable.

LettyYeti (changes requested):

LettyYeti (approved):

PR #992, opened 2022-11-23
Send public key to remote DB

archiepeterjarvis:


archiepeterjarvis:
Tested on rig by going through serial send procedure, public keys show in DB

src/asmcnc/apps/systemTools_app/screens/calibration/screen_serial_numbers.py
```diff
@@ -173,7 +173,7 @@ class UploadSerialNumbersScreen(Screen):
 
     poll_for_end_of_upload = None
 
-    dev_mode = False
+    dev_mode = True
```
LettyYeti, line 176: dev mode should be off

src/asmcnc/apps/systemTools_app/screens/calibration/screen_serial_numbers.py
```diff
@@ -364,6 +364,8 @@ def get_truncated_fw_version(self, version):
 
     def download_and_upload_LB_cal_data(self):
 
+        self.send_public_keys()
```
LettyYeti, line 367: what happens if this fails? will it crash?

src/asmcnc/apps/systemTools_app/screens/popup_system.py
```diff
@@ -384,8 +384,8 @@ def __init__(self, screen_manager, localization):
         ok_string = self.l.get_bold('Ok')
 
         def check_password(*args):
-            if textinput.text == "fac":
-                self.systemtools_sm.open_factory_settings_screen()
+            # if textinput.text == "fac":
```
LettyYeti, line 387: should not be commented out!

LettyYeti (changes requested):
See comments

Also can you please add some integration tests (in unit test format) for the new queries in calibration db?

src/asmcnc/apps/systemTools_app/screens/calibration/screen_serial_numbers.py
```diff
@@ -403,6 +407,22 @@ def report_info_back_to_user_and_return(self, dt):
...
+            response = output.communicate()[0]
+
+            if len(response) == 0:
+                PopupNoSSHFile()
+                return
+
+            self.calibration_db.send_ssh_keys(self.console_serial_input.text, response)
+        except:
+            log("Couldn't find public key file")
+            PopupNoSSHFile()
```
LettyYeti, line 424: Won't this also fail if WiFi connection is poor or if the DB server is down? Can error messages please be updated accordingly because otherwise (e.g. if the WiFi drops out in future) I will definitely forget and get confused when I can find a perfectly valid key file 😅

Also, what happens if the key file is there but there's nothing in it? (e.g. someone (me) has been checking the key file and they've accidentally deleted the contents in the process 😳)

LettyYeti (changes requested):

LettyYeti (approved):

PR #999, opened 2022-11-28
ShortBench graphic

LettyYeti:
See changes in src/asmcnc/skavaUI/widget_virtual_bed.py - everything else is merged from an earlier PR.

Tested on rig with changes made to dollar settings (checked both standard and shorty). 

Screenshot 2022-11-28 at 18 46 02


src/asmcnc/skavaUI/widget_virtual_bed.py
```diff
@@ -110,13 +113,36 @@ class VirtualBed(Widget):
...
+    x_pos_modifier = NumericProperty()
+
     def __init__(self, **kwargs):
     
         super(VirtualBed, self).__init__(**kwargs)
         self.m=kwargs['machine']
         self.sm=kwargs['screen_manager']
+        self.set_up_virtual_bed(0)
+
+    def set_up_virtual_bed(self, dt):
```
DennisKiselev, line 127: When we define functions that are called by clock functions, I think it would make sense to put dt=0 as the parameter so that we don't have to pass in the 0 when we manually call the function

LettyYeti: Have made the change and tested on the rig :)

PR #1003, opened 2022-12-05
Make holding scroll button scroll indefinitely

DennisKiselev:
Holding down the gcode scroll button on job recovery line selection screen now scrolls continuously after a short delay

Tested on windows by spamming buttons a lot

### DIFF
--- a/src/asmcnc/skavaUI/screen_job_recovery.py
+++ b/src/asmcnc/skavaUI/screen_job_recovery.py
@@ -85,7 +85,8 @@ Builder.load_string("""
 
                             Button:
                                 background_color: [0,0,0,0]
-                                on_press: root.scroll_up()
+                                on_press: root.start_scrolling_up()
+                                on_release: root.stop_scrolling_up()
                                 BoxLayout:
                                     padding: 0
                                     size: self.parent.size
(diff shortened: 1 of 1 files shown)


src/asmcnc/skavaUI/screen_job_recovery.py
```diff
@@ -329,16 +331,56 @@ def on_pre_enter(self):
...
             self.selected_line_index -= 1
             self.update_display()
 
-    def scroll_down(self):
+    def scroll_up(self, dt=0):
+        if self.scrolling_up:
+            self.do_scroll_up()
+
+            # Keep scrolling until button released
+            self.scroll_up_event = Clock.schedule_once(self.scroll_up, 0.03)
```
LettyYeti, line 353: I recommend declaring scroll_up_event and scroll_down_event at the top of the class as: 

scroll_up_event = None
scroll_down_event = None

just in case there's a freak situation where the stop function unschedules the event before it's been created - this is rare but I have seen it happen in other cases, and it will cause a crash. 

Then when unscheduling these events, do:

if event != None: Clock.unschedule(event)

LettyYeti (commented):
Also, be sure to test this on the rig - differences in the touch responsivity of the console interface & differences in error in the kivy clock on the different processors may create a different feel of the scroll behaviour. Something that feels smooth on windows might feel janky on the console, and vice versa.

PR #1009, opened 2022-12-13
Version detection

DennisKiselev:
SmartBench model is now worked out using a function from router_machine rather than read from a file

Tested using unit tests, and functionally tested on windows

### DIFF
--- a/src/asmcnc/apps/systemTools_app/screens/screen_build_info.py
+++ b/src/asmcnc/apps/systemTools_app/screens/screen_build_info.py
@@ -644,11 +644,10 @@ class BuildInfoScreen(Screen):
             self.show_more_info.opacity = 1
 
     def get_smartbench_model(self):
-        try:
-            file = open(self.smartbench_model_path, 'r')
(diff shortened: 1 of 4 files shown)


LettyYeti (changes requested):
Logic for the smartbench_model() function looks good, but there's a mistake in the last commit where you forgot to uncomment the platform checks. 

HOWEVER - this is good, because it highlights a problem with these functions. We need to be able to run unit tests automatically, without commenting things in and out (that kind of change sort of negates the point of unit testing).

I suggest adding a default variable to the functions, something like "bypass_platform=False", that can then be accessed from the unit tests for test scenarios. 

Then probs set up the platform line so it does something like 

```
if sys.platform == 'win32' or sys.platform == 'darwin' or bypass_platform: 
    return False
```

(Maybe a more elegant solution would be one platform getter which we can then mock, but that seems like it may be more work to implement? Maybe/maybe not)

Also if you have a better solution then please do that - this was the first thing that came to mind :) 

After this change do a quick HW test as well to make sure everything works as it should :)

LettyYeti (approved):

PR #1010, opened 2022-12-15
Up down jig popups

DennisKiselev:
Up down jig has two new popups:

- When calibrate button is pressed, a popup is shown to tell user to remove the belt before calibration
- When phase one ends, a popup is shown that warns the user before phase two starts

image

image


DennisKiselev:
Updated phase two popup

image

PR #1012, opened 2023-01-05
Debug current setting: router_machine function and factory screen

LettyYeti:
- Checked what wait times need to be for current setting, and removed sleep commands from current setting function
- Set up current setting function to take in arguments for both individual motors and axes
- Add unit tests for current setting function
- Set wait popup to auto-dismiss - this means that folks can click away from it to clear it if they get stuck :) 
- Add status bar that shows motor load values to current adjustment screen 
- Add text box that shows status of protocol commands being sent to serial 
- Ensure TMC registers are read back in when parameters are stored
- Add text boxes that can be used instead of up-down buttons for setting currents, and ensure that current is set when focus goes off of text box
- Add wait pop-ups while currents are being stored or reset


src/asmcnc/skavaUI/popup_info.py
```diff
@@ -746,7 +746,7 @@ def __init__(self, screen_manager, localization, description = False):
                       content=layout_plan,
                       size_hint=(None, None),
                       size=(500, 200),
-                      auto_dismiss= False
+                      auto_dismiss= True
```
DennisKiselev, line 749: Why did you make this change?

LettyYeti: Sometimes the wait popups can get stuck (if something goes wrong - there used to be one with the software update that wouldn't clear if the SW was already on the latest version I think). I wanted to put an "X" in the corner but I couldn't find a straightforward way to do that (maybe a float layout would do it but I'm not sure). I did however find this auto-dismiss parameter that will clear the popup if the user presses on space that isn't the popup. This won't clear popups that are locked on by active processes that cause the UX to lock up (e.g. mid SW updates or downloading logs), but it is useful for cases where something has prevented the wait popup from clearing. 

There was a ticket in the backlog for this anyway (which I've since updated) and it was a quick change I could make while I was putting in the wait popup anyway :)

LettyYeti:
I've realised that currents & raw SG values are not being reset properly if SB is not Idle - I want to add in this "safety"/"error handling" before I merge. Might do it on a separate branch though.

src/asmcnc/skavaUI/popup_info.py
DennisKiselev: Seems logical, I only highlighted this because I can imagine a user clearing a wait popup that shouldn't be cleared before its time has run out which could mess with functionality, such as the wait popup before the nudge screen that I've used in job recovery. In theory a user wouldn't do this so it might be fine but it just didn't feel completely robust. What do you think?

DennisKiselev: On the other hand this may help the user avoid getting locked into a wait popup in job recovery in case some weird bug happens with the machine state... might be fine actually

LettyYeti: I understand, I had the same thought. I personally think that users are unlikely to click out of it most of the time, and it mainly acts as a "safety" if there are issues for any reason.  I think this way is the lesser of two evils. 

That said, if you think that it would be better to not have the auto-dismiss, then I can change it back and we'll keep everything as is :)

DennisKiselev: I think it's probably fine :)

PR #1018, opened 2023-01-14
Maintenance brush times

DennisKiselev:
Maintenance brush times are now received from SC2 spindle if $51 is enabled, and reset functionality also uses SC2 spindle. Checks are used to make sure that spindle is connected in this case.

Tested on rig


src/asmcnc/apps/maintenance_app/widget_maintenance_brush_use.py
```diff
@@ -149,7 +150,31 @@ def __init__(self, **kwargs):
         self.update_strings()
         
     def restore(self):
```
LettyYeti, line 152: I think these functions could be refactored and broken up a bit into: 
- a function (or sequence of functions) that read the info from the spindle (and this could probs go into router_machine)
- UX side
That would make it a little bit more readable, and mean that the read in function was available in router_machine if it was needed elsewhere (e.g. as part of the prompt to replace brushes)

LettyYeti (changes requested):
Few comments on this one, I also want to check what the delays need to be with Boris (or maybe this is something to try experimenting with) because I suspect that they can be tightened up a bit :)

LettyYeti:
From Boris: 

"digital spindle info" data is read from spindle every 20ms. Taking into account processing delays and potential spindle boot lag anything higher than 0.1s should work

(My question was:
do you have any wisdom regarding how long the console should wait between:

1. Turning on the spindle with M3 S0
2. Getting digital spindle info
3. Reading in spindle info
4. Turning off spindle (M5)

Particularly between step 1 and 2))

So I think it's fine to shorten delays - just make sure the console has enough time to read in the ouput from the FW as well :)

src/asmcnc/apps/maintenance_app/widget_maintenance_brush_use.py
DennisKiselev: Had a go at refactoring this but it doesn't look much more readable to me, does it look right to you?

src/asmcnc/apps/maintenance_app/widget_maintenance_brush_save.py
```diff
@@ -165,3 +176,40 @@ def save(self):
...
+        self.brush_reset_test_count = 0
+        self.attempt_reset()
+
+    def attempt_reset(self):
+        def read_info(dt):
+            self.m.s.write_protocol(self.m.p.GetDigitalSpindleInfo(), "GET DIGITAL SPINDLE INFO")
+            Clock.schedule_once(get_info, 1)
+
+        def get_info(dt):
+            self.initial_run_time = self.m.s.spindle_brush_run_time_seconds
```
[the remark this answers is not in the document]
DennisKiselev: They work a bit differently so the code can't be shared between these screens

src/asmcnc/apps/maintenance_app/widget_maintenance_brush_use.py
DennisKiselev: Nevermind, after testing it seems that the refactoring just breaks the functionality and I can't work out why, I think it would be best to just leave it as is. Probably best not to split wait popups up from the functionality that they are waiting for, unless there's a smarter method than what I was doing

LettyYeti: Ok no worries, well done for having a stab at it anyway :) 

There probably is a solution, but if it's not easy to find let's leave it for now.

LettyYeti (changes requested):
Only thing we need now are the final approvals on popup strings & the translations, everything else looks good though :)

PR #1030, opened 2023-02-09
Auto set setting 51

DennisKiselev:
On the spindle cooldown settings tab in the maintenance app, selecting a spindle from the dropdown and saving now sets the $51 setting depening on whether or not the spindle is SC2. Additionally, the option to select an SC2 spindle is only given when the PRO+ unlock file and setting 51 both exist.

$51 command is sent every time, regardless of whether it needs changing or not - Boris confirmed that this is fine, as EEPROM wear happens: 'Only when writing different value. This applies to most EEPROMs in the market.
Atmega EEPROM "has an endurance of at least 100,000 write/erase cycles" (from datasheet)
"Write" means wirting "0" in one bit,  "Erase" means write "1"s to all bits in the byte or multiple bytes
Meaning that writing $51=1 when $51 is 0 will cause mupltiple bytes erase and rewrite.'

Tested on rig

### DIFF
--- a/src/asmcnc/apps/maintenance_app/screen_maintenance.py
+++ b/src/asmcnc/apps/maintenance_app/screen_maintenance.py
@@ -581,7 +581,10 @@ class MaintenanceScreenClass(Screen):
         # Only show SC2 options if machine supports it
         self.spindle_settings_widget.spindle_brand.values = self.spindle_settings_widget.brand_list_sc1
         try:
-            if os.path.exists('/home/pi/proplus.txt') and self.m.s.setting_51:
+            # Check if $51 exists
+            self.m.s.setting_51
+            # Check if PRO+ is activated
+            if os.path.exists('/home/pi/proplus.txt'):
                 self.spindle_settings_widget.spindle_brand.values = self.spindle_settings_widget.brand_list_sc2
         except:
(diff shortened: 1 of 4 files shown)


src/asmcnc/apps/maintenance_app/screen_maintenance.py
```diff
@@ -576,6 +578,17 @@ def on_pre_enter(self):
...
         self.spindle_settings_widget.rpm_override = self.m.spindle_cooldown_rpm_override
 
+        # Only show SC2 options if machine supports it
+        self.spindle_settings_widget.spindle_brand.values = self.spindle_settings_widget.brand_list_sc1
+        try:
+            # Check if $51 exists
+            self.m.s.setting_51
+            # Check if PRO+ is activated
+            if os.path.exists('/home/pi/proplus.txt'):
+                self.spindle_settings_widget.spindle_brand.values = self.spindle_settings_widget.brand_list_sc2
```
LettyYeti, line 588: using append() might have worked better here than having two lists (provided the kivy dropdown values allows it), but it doesn't need to block merge :)

LettyYeti (approved):
Looks good, ready for merge :)

PR #1031, opened 2023-02-09
Maintenance brush times feedback

DennisKiselev:
Some changes to PRO+ maintenance brush times based on feedback:

- Don't show settings saved popup if reset didn't work
- Don't allow user to input brush use values for SC2 - can only use reset or restore buttons to change input values
- Change new error messages to existing localised ones while we wait for finalised localised ones

### DIFF
--- a/src/asmcnc/apps/maintenance_app/widget_maintenance_brush_save.py
+++ b/src/asmcnc/apps/maintenance_app/widget_maintenance_brush_save.py
@@ -88,7 +88,7 @@ class BrushSaveWidget(Widget):
 
         # Machine must be idle if a spindle brush reset is attempted
         if not self.m.state().startswith('Idle'):
-            popup_info.PopupError(self.sm, self.l, self.l.get_str("Please ensure the machine is idle before attempting to save."))
+            popup_info.PopupError(self.sm, self.l, self.l.get_str("There was a problem saving your settings."))
             return
 
         # TIME FOR DATA VALIDATION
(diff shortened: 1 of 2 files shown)


src/asmcnc/apps/maintenance_app/widget_maintenance_brush_save.py
```diff
@@ -177,6 +168,25 @@ def save(self):
             popup_info.PopupError(self.sm, self.l, warning_message)
             return
 
+    def write_new_values(self):
+    
+        # write new values to file
+        if self.m.write_spindle_brush_values(self.use, self.lifetime):
```
LettyYeti, line 174: We only want brush use times to be saved to the console for SC1, NOT SC2

LettyYeti (changes requested):
We ONLY want to save spindle brush usage to the console for SC1s, and NOT when using SC2s. Lifetime needs saving for both scenarios though :)

src/asmcnc/apps/maintenance_app/widget_maintenance_brush_save.py
```diff
@@ -197,13 +209,14 @@ def check_info(dt):
                 if self.brush_reset_test_count == 5:
                     self.m.s.write_command('M5')
                     self.wait_popup.popup.dismiss()
-                    popup_info.PopupError(self.sm, self.l, self.l.get_str("Could not reset brush timer! Please try again."))
+                    popup_info.PopupError(self.sm, self.l, self.l.get_str("There was a problem saving your settings."))
                 else:
                     self.attempt_reset()
             else:
                 self.m.s.write_command('M5')
                 self.sm.get_screen('maintenance').brush_monitor_widget.update_percentage()
```
LettyYeti, line 217: this line gets called in write_new_values as well - no need to duplicate :)

PR #1036, opened 2023-02-14
Read info from spindle when using SC2

DennisKiselev:
Changed go screen to use GET_SPINDLE_INFO on entry if SC2 is being used, in order to decide whether to show brush use popup, rather than using stored values.

Tested on rig, and windows for edge cases


src/asmcnc/skavaUI/screen_go.py
```diff
@@ -463,70 +463,95 @@ def on_pre_enter(self, *args):
...
+        self.m.s.write_command('M5')
+        self.wait_popup.popup.dismiss()
+        # If info was not obtained successfully, spindle production year will equal 99
+        if self.m.s.spindle_production_year != 99:
+            try: # Just in case of weird errors
+                self.check_brush_use_and_lifetime(self.m.s.spindle_brush_run_time_seconds, self.m.spindle_brush_lifetime_seconds)
+                return
+            except:
+                pass
+        popup_info.PopupError(self.sm, self.l, self.l.get_str("Error!"))
```
DennisKiselev, line 490: Needs an actual message!!!

src/asmcnc/skavaUI/screen_go.py
```diff
@@ -463,70 +463,95 @@ def on_pre_enter(self, *args):
...
-                        ).replace("N00", time_since_lubricated_string) + \
-                        "\n\n" + \
-                        self.l.get_str("Will you lubricate the Z head now?") + "\n\n" + \
-                        self.l.get_str("Saying 'OK' will reset this reminder.")
-                )
-
-                lubrication_reminder_popup = popup_info.PopupReminder(self.sm, self.am, self.m, self.l,
-                                                                      lubrication_warning, 'lubrication')
+            # Statement below is evaluated left to right, so fw version has to be >= 2.2.8 for setting_51 to be checked
+            if self.m.is_machines_fw_version_equal_to_or_greater_than_version('2.2.8', 'Spindle lifetime check') and self.m.theateam() and self.m.s.setting_51:
```
LettyYeti, line 467: swap self.m.s.setting_51 for new [get_dollar_setting() function](https://github.com/YetiTool/easycut-smartbench/blob/415e3aaa27eef0745dd4450f192af62464fd989b/src/asmcnc/comms/router_machine.py#L1396), just protects in case setting_51 doesn't exist

LettyYeti (changes requested):

PR #1043, opened 2023-02-23
scrape_last_feed_command and get_is_constant_feed_rate for YetiPilot

LettyYeti:
- Slight refactor of functions pulled from YP branch:
    - Used a variation of Dennis's code for scraping feedrate (this scrapes a float rather than an int)
    - ensured args were passed to functions to keep them more general, and allow ease of testing
    - moved functions out of YP into other parts of the codebase where they can be used more generally (e.g. the feedrate scrape can by called by job recovery in future) 
    - moved location of scrape_last_feed_command in job data module in the hope that it's less likely to create merge conflict with job recovery later on

- Unit tests to ensure that functionality is good and working as expected :)

Tested via unit test, and by running main.py on computer

### DIFF
--- a/src/asmcnc/comms/router_machine.py
+++ b/src/asmcnc/comms/router_machine.py
@@ -1428,6 +1428,11 @@ class RouterMachine(object):
 # SPEED AND FEED GETTERS
     def feed_rate(self): return int(self.s.feed_rate)
 
+    def get_is_constant_feed_rate(self, feed_override_percentage, feed_rate, current_line_number):
+        last_modal_feed_rate = self.jd.scrape_last_feed_command(self.job_gcode_running, current_line_number)
+        constant_feed_target = last_modal_feed_rate * feed_override_percentage / 100
+        return abs(constant_feed_target - feed_rate) < 50, last_modal_feed_rate
+
     def spindle_speed(self): 
         if self.spindle_voltage == 110: 
             # if not self.spindle_digital or not self.fw_can_operate_digital_spindle(): # this is only relevant much later on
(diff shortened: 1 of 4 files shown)


src/asmcnc/job/job_data.py
```diff
@@ -166,6 +166,20 @@ def set_job_filename(self, job_path_and_name):
         else:
             self.job_name = self.filename.split("/")[-1]
 
+
+    def scrape_last_feed_command(self, job_gcode_object, index): 
+
+        try:
+            feedrate_line = next((s for s in reversed(job_gcode_object[:index]) if 'F' in s), None)
+            if feedrate_line:
+                feedrate = re.match('\d+(\.\d+)?$',feedrate_line[feedrate_line.find("F")+1:]).group()
```
DennisKiselev, line 175: Is it possible that the feedrate is defined in the middle of the line rather than the end? If this is the case then the $ will stop it from matching since it's not the end of the line. Seems unlikely but worth checking

LettyYeti: Good spot @DennisKiselev, thank you! Have updated unit tests and removed $ sign :) 

Please check and resolve this conversation if you're happy 😁

src/asmcnc/comms/router_machine.py
```diff
@@ -1428,6 +1428,10 @@ def g28_z(self): return float(self.s.g28_z)
 # SPEED AND FEED GETTERS
     def feed_rate(self): return int(self.s.feed_rate)
 
+    def get_is_constant_feed_rate(self, last_modal_feed_rate, feed_override_percentage, current_feed_rate):
```
archiepeterjarvis, line 1431: this function isn't very general, but it can live here

LettyYeti: Even though it's only used by YP at the moment, it's not dependent on other parts of the YP algorithm, and I think it could have uses outside of YP in future if we wanted :)

PR #1051, opened 2023-02-27
YP Container & SC2 prompts -> on_pre_enter

LettyYeti:
- Move container show/hide into on_pre_enter - this prevents visible "jumps" on the screen
- Move start of SC2 data check to on_pre_enter - this creates the illusion of it being faster 
- New "is_using_sc2" func in router_machine
- Ensure YP toggle is disabled when the container is hidden - otherwise it is possible to click on it still!

### DIFF
--- a/src/asmcnc/skavaUI/screen_go.py
+++ b/src/asmcnc/skavaUI/screen_go.py
@@ -513,7 +513,7 @@ class GoScreen(Screen):
 
     def get_spindle_info(self, dt):
         self.m.s.write_protocol(self.m.p.GetDigitalSpindleInfo(), "GET DIGITAL SPINDLE INFO")
-        Clock.schedule_once(self.read_spindle_info, 0.8)
+        Clock.schedule_once(self.read_spindle_info, 1)
 
     def read_spindle_info(self, dt):
         self.m.s.write_command('M5')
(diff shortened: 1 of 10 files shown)


src/asmcnc/skavaUI/screen_go.py
```diff
@@ -484,32 +490,41 @@ def on_pre_enter(self, *args):
...
 
+        if self.show_maintenance_prompts() and use_sc2: self.get_sc2_brush_data()
+
     def on_enter(self):
 
-        # Check for SC2 compatability
-        if self.m.is_machines_fw_version_equal_to_or_greater_than_version('2.2.8', 'Spindle lifetime check') \
-            and self.m.theateam() and self.m.get_dollar_setting(51) and self.m.stylus_router_choice != 'stylus':
-            sc2_compatible = True
+        if not self.getting_brush_data and self.show_maintenance_prompts():
```
DennisKiselev, line 497: I think this should also have the condition `not self.m.is_using_sc2()`

DennisKiselev: Also, the `self.getting_brush_data` doesn't seem to have a use as far as I can see - I would move this to on_pre_enter and rewrite this as:
`if self.show_maintenance_prompts():
    if use_sc2: self.get_sc2_brush_data()
    else: check_brush_use_and_lifetime`
Let me know if I'm missing something here

LettyYeti: For some reason I thought the popups had to be called on_enter, but I have no idea why.

Anyway, moved everything into pre-enter and retested :). Please re-review.

Thank you!

PR #1061, opened 2023-03-01
Add text for yetipilot spindle data loss error screen

DennisKiselev:
Added frontend for information for pause screen when job is paused by yetipilot due to spindle data loss. Additionally added popup for this screen that contains a QR code, and updated previous yetipilot pause screen to also have this QR code.

Tested on windows with unit test

image
image


src/asmcnc/skavaUI/screen_stop_or_resume_decision.py
```diff
@@ -133,17 +133,34 @@ def __init__(self, **kwargs):
...
+                self.l.get_bold('Resume') + '[b]' +  " (>)" +'[/b]' + \
+                "\n" + \
+                self.l.get_str("Pressing resume will continue the job from the point at which it was paused.") + \
+                "\n\n" + \
+                self.l.get_bold('Scan the QR code to learn more about this error.') + \
+                "\n" + \
+                self.l.get_bold("Or visit <URL>").replace('<URL>', 'www.yetitool.com/support > Knowledge Base')
+            )
+
+            popup_info.PopupQRInfo(self.sm, self.l, 500, info, "./asmcnc/skavaUI/img/qr_yetipilot_info.png")
```
LettyYeti, line 164: This can be refactored to have the cancel_resume text block just set up once, and then the rest appended depending on need. Doesn't need to block merge though :)

PR #1068, opened 2023-03-04
Stop auto resume after unpause

DennisKiselev:
When resuming a job after a pause, the job no longer automatically resumes, and instead the user is now required to press the resume button on the go screen to allow them time to adjust settings.

Tested on rig

### DIFF
--- a/src/asmcnc/skavaUI/screen_go.py
+++ b/src/asmcnc/skavaUI/screen_go.py
@@ -655,6 +655,11 @@ class GoScreen(Screen):
             else:
                 self.m.resume_after_a_stream_pause()
                 self.start_or_pause_button_image.source = "./asmcnc/skavaUI/img/pause.png"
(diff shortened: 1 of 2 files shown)


src/asmcnc/skavaUI/screen_go.py
```diff
@@ -649,7 +649,11 @@ def start_or_pause_button_press(self):
 
         log('start/pause button pressed')
         if self.is_job_started_already:
-            self._pause_job()
+            if not self.m.is_machine_paused:
+                self._pause_job()
```
LettyYeti, line 653: probably better to put the button image change here instead of in the pause screen, makes it more general and keeps it local to just the go screen

LettyYeti (changes requested):

LettyYeti (approved):

PR #1077, opened 2023-03-06
add funcs from job_data and hook in

archiepeterjarvis:
tested on rig

### DIFF
--- a/src/asmcnc/job/job_data.py
+++ b/src/asmcnc/job/job_data.py
@@ -423,13 +423,7 @@ class JobData(object):
         return self.g0_lines
 
     def get_spindle_speeds(self, gcode):
-        self.spindle_speeds[:] = []
-        for i, line in enumerate(gcode):
-            if 'S' in line:
-                s_index = line.index('S')
-                end_index = s_index + 1
-                while end_index < len(line) and not line[end_index].isalpha():
-                    end_index += 1
-                s_value = float(line[s_index + 1:end_index].strip())
-                self.spindle_speeds.append([i, s_value])
+        pattern = re.compile(r"S\d+\.?\d*")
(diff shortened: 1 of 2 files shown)


src/asmcnc/job/job_data.py
```diff
@@ -405,3 +409,27 @@ def post_job_data_update_post_send(self):
...
+            if 'G0' in line:
+                self.g0_lines.append(i)
+                for j, next_line in enumerate(gcode[i + 1:]):
+                    if 'G1' in next_line or 'G2' in next_line or 'G3' in next_line:
+                        break
+                    else:
+                        self.g0_lines.append(i + j + 1)
+        return self.g0_lines
+
+    def get_spindle_speeds(self, gcode):
```
LettyYeti, line 425: when does this get called? Has potential to be slow and locking bc it's a for loop

src/asmcnc/job/job_data.py
```diff
@@ -405,3 +409,27 @@ def post_job_data_update_post_send(self):
         self.post_production_notes = ''
         self.batch_number = ''
         self.percent_thru_job = 0
+
+    def get_excluded_line_numbers(self, gcode):
```
LettyYeti, line 413: Can this function either be renamed or commented with details of what it's doing. Also, when does it run, as it's a for loop and has potential to lock up kivy

LettyYeti (changes requested):

src/asmcnc/job/job_data.py
archiepeterjarvis: renamed

src/asmcnc/job/job_data.py
archiepeterjarvis: inside run_job in serial_connection

PR #1082, opened 2023-03-08
Scrape GRBL mode data as lines get sent to GRBL buffer, and monitor top of tracking list for modes

LettyYeti:
New grbl_mode_tracker list in job_data module stores tuples of (grbl motion mode, feed rate, spindle speed). This is added to each time a line is written through the serial port. 

When a new line is read in from the grbl status, it deletes the top of the tracker; this means that the length of the tracker is kept just between the last line sent and the last line executed, and ensures that the current grbl mode is always at the top of the list. 

Tested via: 
- unit tests
- on rig (tracking length of tracker against difference between sent lines and executing lines)
- endurance test on rig
- running/pausing/completing jobs on rig
- as part of yetipilot


src/asmcnc/comms/serial_connection.py
```diff
@@ -519,60 +517,96 @@ def _reset_counters(self):
...
+    ## GCODE BUFFER STUFFING TRACKING
+    last_line_executed = 0
+
+    last_sent_motion_mode = ""
+    last_sent_feed = 0
+    last_sent_speed = 0
+
+    feed_pattern = re.compile(r"F\d+\.?\d*")
+    speed_pattern = re.compile(r"S\d+\.?\d*")
+    g_motion_pattern = re.compile(r"G([0-3])(\D|$)")
```
DennisKiselev, line 551: I think gcodes like G00 and G01 are valid, so maybe change to G0?([0-3])(\D|$)

LettyYeti: G0?([0-3])(\D|$) doesn't seem to work correctly (it's fine when I test it on https://regex101.com/ but not with python 2.7, so I suspect that there's some difference there), as it returns the "0" as the first group. Investigating now...

LettyYeti:
Re-tested with unit tests (amended) and through job run on rig

PR #1103, opened 2023-03-20
Only increment job time when job is running

archiepeterjarvis:


archiepeterjarvis:
Tested on rig, started job, cancelled, paused, etc. Works as expected.

PR #1106, opened 2023-03-22
Job recovery nudge button

DennisKiselev:
Added a button to nudge screen which can be pressed to set the datum after performing a nudge. This replaces the existing way of setting the datum by proceeding to the next screen. Additionally made the big popups on recovery and nudge screen scrollable to fit text in more nicely.

Tested on rig, localisation checked on windows

image
image
image


src/asmcnc/comms/foreign_dictionary.txt
```diff
@@ -771,6 +771,11 @@ The toolpiece should lightly touch the edge of the cut path.	Das Werkzeug sollte
...
 SSH	SSH	SSH	Shell Sicura	SSH	SSH	SSH	SSH
+Once you have nudged your X and Y axis, press the "SET" button to save your new datum.	Wenn Sie Ihre X- und Y-Achse justiert haben, drücken Sie die „SET“-Taste, um den neuen Startpunkt zu speichern.	Après avoir déplacé vos axes X et Y, appuyez sur le bouton "SET" pour enregistrer votre nouveau point de référence.	Una volta che hai accostato gli assi X e Y, premi "SET" per salvare la nuova origine.	Kun olet siirtänyt X- ja Y-akselit, paina "SET"-painiketta uuden datan tallentamiseksi.	Zodra u uw X- en Y-as hebt ingesteld, drukt u op de toets "SET" om uw nieuwe referentiepunt op te slaan.	Po dokonaniu mikroprzesunięć w osiach X i Y, naciśnij przycisk "SET", aby zapisać nową pozycję. 	Når du har rykket din X- og Y-akse, tryk da på ”SET”-knappen for at gemme dine nye data.
```
LettyYeti, line 774: Ensure that spreadsheet key is updated to reflect additions here

src/asmcnc/skavaUI/screen_nudge.py
```diff
@@ -257,5 +252,22 @@ def generate_gcode():
...
+
+        if abs(self.diff_x) > 3 or abs(self.diff_y) > 3:
+            # Display overall distance nudged to user, to 2dp
+            nudge_distance = "{:.2f}".format(math.hypot(self.diff_x, self.diff_y))
+            popup_nudge.PopupNudgeWarning(self.sm, self.m, self.l, nudge_distance)
+        else:
+            popup_nudge.PopupNudgeDatum(self.sm, self.m, self.l)
+
+    def set_datum(self):
+        # G54 and initial pos values do not get updated while on this screen, so nudges are always calculated from the same point
```
LettyYeti, line 267: I don't fully understand this comment/how this works? Are the g54 values not updated when the G10 command is written?

src/asmcnc/skavaUI/screen_job_recovery.py
```diff
@@ -476,18 +476,21 @@ def next_screen(self):
...
             self.jd.job_recovery_selected_line = self.selected_line_index + 1
             self.go_xy()
-            Clock.schedule_once(self.proceed_to_next_screen, 0.4)
+            Clock.schedule_once(self.wait_for_idle, 0.4)
         else:
             error_message = self.l.get_str('Please ensure machine is idle before continuing.')
             popup_info.PopupError(self.sm, self.l, error_message)
 
-    def proceed_to_next_screen(self, dt):
+    def wait_for_idle(self, dt):
```
LettyYeti, line 484: probably need some kind of break condition or event handling here; if there is an alarm or door or something, then that screen will come up, but the clocks here will continue in the background and once the machine state goes to idle then it will bring up the nudge screen, potentially at a surprising point :)

src/asmcnc/comms/router_machine.py
```diff
@@ -1538,15 +1538,30 @@ def wait_for_movement_to_complete(dt):
...
+
+        if x is not None:
+            datum_command += " X" + str(x)
+        if y is not None:
+            datum_command += " Y" + str(y)
+        if z is not None:
+            datum_command += " Z" + str(z)
+
+        self.s.write_command(datum_command)
+        self.get_grbl_status()
```
LettyYeti, line 1562: Might want to double check that this gives grbl enough time to change the datum & get the status - Boris may be able to advise. It may be worth sending these commands as a sequential stream so that the next command doesn't send until the oks have come back

LettyYeti (changes requested):

src/asmcnc/comms/router_machine.py
DennisKiselev: The old Z datum setter function did it this way, so it should work right? At least as far as I've tested it works fine. I tried sequential stream but it messes with the LED command that the existing functions use afterwards

LettyYeti: great, no worries :)

LettyYeti (approved):

PR #1113, opened 2023-03-27
Break up text on line selection screen popup

DennisKiselev:
Already added scrollable popups in nudge button pull request - this is just for the main screen popup

image


src/asmcnc/skavaUI/screen_job_recovery.py
```diff
@@ -479,7 +479,7 @@ def get_info(self):
                self.l.get_bold("Ensure that you recover the job from a point where it was running normally, and SmartBench's position was accurate.") + ' ' + \
```
LettyYeti, line 479: Add break here

LettyYeti: It won't let me add comments where there's no edit, but also add break after:

Arc movements (G2 and G3) may cause the software to think that the job failed earlier than it did.

PR #1114, opened 2023-03-28
Inrush prevention

archiepeterjarvis:
Wait 20 statuses before allowing feed adjustments to prevent inrush slowing the job down.

### DIFF
--- a/src/asmcnc/comms/serial_connection.py
+++ b/src/asmcnc/comms/serial_connection.py
@@ -382,7 +382,7 @@ class SerialConnection(object):
                         if self.digital_spindle_ld_qdA >= 0 \
                                 and self.grbl_ln is not None \
                                 and self.digital_spindle_mains_voltage >= 0 \
-                                and self.inrush_counter == 12:
+                                and self.inrush_counter == 20:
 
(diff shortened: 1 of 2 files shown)


archiepeterjarvis:
Tested on rig

src/asmcnc/comms/serial_connection.py
```diff
@@ -910,9 +912,13 @@ def process_grbl_push(self, message):
             if not re.search(self.digital_load_pattern, message):
                 self.inrush_counter = 0
 
-            elif self.inrush_counter <= 11:
+            elif self.inrush_counter < 20:
```
LettyYeti, line 915: did you intend to put self.inrush_max here?

src/asmcnc/comms/serial_connection.py
```diff
@@ -783,6 +783,8 @@ def update_machine_runtime(self):
     # Spindle data "inrush" counter
     digital_load_pattern = re.compile(r"Ld:\d+,\d+,\d+,\d+")
     inrush_counter = 0
+    inrush_max = 20
+    in_inrush = False
```
LettyYeti, line 787: where does this get set to true?

PR #1119, opened 2023-03-29
Added spindle health check to maintenance app

LettyYeti:
Added + tab to the maintenance app, currently does not show in any case due to this: 

`        if not self.m.theateam() or True:
            self.update_strings()
            return`

so that can be changed once functionality is in :) 

Tested with the screen test & on rig


src/asmcnc/apps/maintenance_app/screen_maintenance.py
```diff
@@ -602,8 +639,13 @@ def on_enter(self):
             self.tab_panel.switch_to(self.laser_tab)
         elif self.landing_tab == 'spindle_tab':
             self.tab_panel.switch_to(self.spindle_tab)
+        elif self.landing_tab and self.m.theateam() == 'spindle_health_check_tab':
```
DennisKiselev, line 642: I think you got this the wrong way around

src/asmcnc/apps/maintenance_app/screen_maintenance.py
```diff
@@ -538,8 +554,29 @@ def __init__(self, **kwargs):
         self.z_lubrication_reminder_widget = widget_maintenance_z_lubrication_reminder.ZLubricationReminderWidget(machine=self.m, screen_manager=self.sm, localization=self.l)
         self.z_lubrication_reminder_container.add_widget(self.z_lubrication_reminder_widget)
 
+        # OPTIONAL + TAB
+
+        ## Enable tab
+
+        if not self.m.theateam() or True:
+            self.update_strings()
+            return
```
DennisKiselev, line 563: I think the better way of doing this would be to invert the condition and nest all of the code below, aside from update_strings. This way the update_strings function doesn't need to be duplicated, and I think using return should be avoided in case you want to add code below this and forget about it

src/asmcnc/apps/maintenance_app/screen_maintenance.py
LettyYeti: oh dear 😅


but somehow it seemed to work ?? 🤯

LettyYeti: ah no, I just don't remember the order in which I tested and added, and then didn't catch it on retesting 🤦‍♀️

LettyYeti:
Also added logic to add the widgets on_pre_enter (if they don't exist), in case an upgrade has happened

tested with test branch to enable upgrade app [here](https://github.com/YetiTool/easycut-smartbench/compare/maintenance_app_setup...test_maintenance_app_after_upgrade?expand=1)

src/asmcnc/apps/maintenance_app/screen_maintenance.py
```diff
@@ -538,6 +554,13 @@ def __init__(self, **kwargs):
         self.z_lubrication_reminder_widget = widget_maintenance_z_lubrication_reminder.ZLubricationReminderWidget(machine=self.m, screen_manager=self.sm, localization=self.l)
         self.z_lubrication_reminder_container.add_widget(self.z_lubrication_reminder_widget)
 
+        # OPTIONAL + TAB AND WIDGETS
+
+        ## Enable tab
+
+        if self.m.theateam(): # or True:
```
DennisKiselev, line 561: Don't forget the True

DennisKiselev: (make it "if self.m.theateam() and False")

PR #1123, opened 2023-03-30
Reorder choices

LettyYeti:
Reorder sequence in which yetipilot options are picked, to start with material type. 

Requires in situ testing


LettyYeti:
Tested by Benji on SB

PR #1130, opened 2023-04-02
Localise health check pause screen

DennisKiselev:
German and Finnish are autotranslated

Tested with unit test in all languages


DennisKiselev:
German and Finnish have been corrected

PR #1131, opened 2023-04-02
Localise plus safety screens

DennisKiselev:
Localisation for safety information screen. Translations for German and Finnish are auto translated, as well as the two warning lines in all languages:
- "The Spindle motor MUST be clamped securely BEFORE plugging in the Spindle motor cables."
- "If you start any job with the Spindle motor health check enabled, your tool MUST be rated up to 24,000 RPM."

Tested with unit test in all languages


DennisKiselev:
German and Finnish have been checked

PR #1157, opened 2023-04-14
Increase default currents to high values, and add in current change for Y axis

LettyYeti:
Set: 
```
    double_stack_single_driver_x_current = 31
    double_stack_dual_driver_x_current = 27

    default_z_current = 31
    default_y_current = 28
```

Added in code to set, check, and report on setting of the Y current as well. 

Tested on Dev A through test specific branch [high_current_ZHQC_test](https://github.com/YetiTool/easycut-smartbench/tree/high_current_ZHQC_test) for use of port. 
Checked with main app that currents were getting changed & reported in current screen was same as set by ZHQC.

### DIFF
--- a/src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up.py
+++ b/src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up.py
@@ -838,7 +838,7 @@ class ZHeadPCBSetUp(Screen):
                 if int(self.m.TMC_motor[motor].standStillCurrentScale) != int(expected_current):
                     outcome_current_correct = False
 
-            def check_temp_coeff(motor, expected_coeff)
+            def check_temp_coeff(motor, expected_coeff):
                 if int(self.m.TMC_motor[motor].temperatureCoefficient) != int(expected_coeff):
                     outcome_screen.thermal_coefficients_correct = False
 
--- a/src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up.py
+++ b/src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up.py
@@ -304,7 +304,6 @@ Builder.load_string("""
                             Label:
                                 id: recommended_z_current_label
                                 size_hint_x: 0.8
-                                text: "25"
                                 text_size: self.size
                                 markup: 'True'
                                 halign: 'left'
@@ -330,7 +329,6 @@ Builder.load_string("""
                             TextInput:
                                 id: other_z_current_textinput
                                 size_hint_x: 0.4
-                                text: "25"
                                 input_filter: "int"
                                 multiline: False
                                 font_size: "22sp"
@@ -444,6 +442,7 @@ class ZHeadPCBSetUp(Screen):
     number_of_drivers = 4
 
     x_current = str(single_stack_dual_driver_x_current)
+    y_current = str(default_y_current)
     z_current = str(default_z_current)
 
     x_thermal_coefficient = str(default_x_thermal_coefficient)
@@ -528,6 +527,7 @@ class ZHeadPCBSetUp(Screen):
         self.recommended_z_current_checkbox.state = "normal"
         self.other_z_current_checkbox.state = "normal"
 
+        self.recommended_z_current_label.text = str(self.default_z_current)
         self.other_z_current_textinput.text = str(self.z_current)
         self.z_current = self.set_value_to_update_to(self.recommended_z_current_label, self.recommended_z_current_checkbox)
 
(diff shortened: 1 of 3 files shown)


src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up.py
```diff
@@ -831,6 +835,11 @@ def check_registers_are_correct():
             if int(self.m.TMC_motor[TMC_X1].standStillCurrentScale) != int(self.x_current): self.sm.get_screen("qcpcbsetupoutcome").x_current_correct = False
             if int(self.m.TMC_motor[TMC_X2].standStillCurrentScale) != int(self.x_current): self.sm.get_screen("qcpcbsetupoutcome").x_current_correct = False
 
+            if int(self.m.TMC_motor[TMC_Y1].ActiveCurrentScale) != int(self.y_current): self.sm.get_screen("qcpcbsetupoutcome").y_current_correct = False
+            if int(self.m.TMC_motor[TMC_Y2].ActiveCurrentScale) != int(self.y_current): self.sm.get_screen("qcpcbsetupoutcome").y_current_correct = False
+            if int(self.m.TMC_motor[TMC_Y1].standStillCurrentScale) != int(self.y_current): self.sm.get_screen("qcpcbsetupoutcome").y_current_correct = False
+            if int(self.m.TMC_motor[TMC_Y2].standStillCurrentScale) != int(self.y_current): self.sm.get_screen("qcpcbsetupoutcome").y_current_correct = False
+
```
archiepeterjarvis, line 842: Could assign var to screen instead of getting each time

src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up_outcome.py
```diff
@@ -190,6 +210,10 @@ def on_enter(self):
         elif str(self.m.s.fw_version).startswith("1"): self.x_current_image.source = self.undetermined_image
         else: self.x_current_image.source = self.fail_image
 
+        if self.y_current_correct: self.y_current_image.source = self.success_image
+        elif str(self.m.s.fw_version).startswith("1"): self.y_current_image.source = self.undetermined_image
+        else: self.y_current_image.source = self.fail_image
+
```
archiepeterjarvis, line 216: PEP8? ;)

src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up.py
```diff
@@ -443,6 +442,7 @@ class ZHeadPCBSetUp(Screen):
     number_of_drivers = 4
 
     x_current = str(single_stack_dual_driver_x_current)
+    y_current = str(default_y_current)
```
archiepeterjarvis, line 445: why are these strings?

src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up.py
```diff
@@ -799,10 +802,11 @@ def set_currents_and_coeffs():
                 self.m.set_thermal_coefficients("Y", int(self.y_thermal_coefficient)) and \
                 self.m.set_thermal_coefficients("Z", int(self.z_thermal_coefficient)) and \
                 self.m.set_motor_current("Z", int(self.z_current)) and \
+                self.m.set_motor_current("Y", int(self.y_current)) and \
```
archiepeterjarvis, line 805: links to comment above about why these are strings

src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up.py
LettyYeti: I set all of the outputs that would be set by screen selection to be strings, as selected values all get passed through this function: 

`    def set_value_to_update_to(self, text_obj, radio_button):
        if radio_button != None: 
            radio_button.state ='down'
            radio_button.active = True
        value_to_set = re.findall('[0-9.]+', text_obj.text)[0]
        return value_to_set`
        
and sometimes the "value_to_set" is a string value (e.g. FW version).

In hindsight I could have wrapped the function call in `int()`, rather than setting these values to be strings, but I'm honestly not sure which is better from a readability perspective? Alternatively, it _could_ return a string only if it can't convert the return value to an int... but that seems worse?

src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up.py
```diff
@@ -667,6 +668,7 @@ def print_settings_to_set(self):
         print("FW version " + str(self.firmware_version))
 
         print("X current: " + str(self.x_current))
+        print("Y current: " + str(self.y_current))
```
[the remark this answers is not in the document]
LettyYeti: this is because I have a horrible tendency to try and concatenate ints...

src/asmcnc/production/z_head_qc_jig/z_head_qc_pcb_set_up.py
archiepeterjarvis: Yep that makes sense. 

I just questioned it as in ser conn a lot of variables are initialised or stored as strings where int/float would be expected.

PR #1161, opened 2023-04-19
close yp popup at job end

archiepeterjarvis:
tested on machine, ran job with popup open, waited for job end.

### DIFF
--- a/src/asmcnc/core_UI/job_go/popups/popup_yetipilot_settings.py
+++ b/src/asmcnc/core_UI/job_go/popups/popup_yetipilot_settings.py
@@ -394,7 +394,7 @@ class PopupYetiPilotSettings(Widget):
             self.popup.dismiss()
 
             if self.sm.has_screen('go'):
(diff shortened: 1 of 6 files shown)


LettyYeti (changes requested):
Currently it won't clear if the job is cancelled by something beyond the user's control, like an alarm. 

For a more general solution it might be better to put this dismiss in the on_leave() from the go screen, and I'd also keep the popup as a "child" of the yp_widget rather than the yp class.

This is to avoid the yp class getting cluttered with UI where possible, and it should make it clear that the popup "belongs" to the widget (I think the only call to open it is in the widget).

It would also help readability if the popup object (currently self.yp.popup) was renamed to be something more specific, e.g. yp_settings_popup. I'm used to seeing just "popup" as the object name inside popup classes, but never outside, and it really threw me! This also reduces the likelihood of multiple popups called from one place all getting called "popup" 😅.

archiepeterjarvis:
after changes: tested on rig, opened popup, waited for job end

LettyYeti (approved):

PR #1167, opened 2023-04-28
Add error message for loading file with over 10 million lines

DennisKiselev:
Added an error message to display to the user when they attempt to load a file with over 10 million lines

Tested on windows, in all languages

image


LettyYeti (approved):
Looks good :)

PR #1168, opened 2023-05-02
Exclude leds

DennisKiselev:
LED commands are now excluded from having line numbers added to them.

Tested using unit tests, and on rig with yetipilot enabled on a file with an LED command on it, ensuring that the LED colour is changed correctly


LettyYeti (approved):
Looks good :)

PR #1172, opened 2023-05-16
V2. Changed the password input text to single line in the Factory app in System tools

LettyYeti:


LettyYeti:
This looks good, just need the comments on how it was tested :)

PR #1173, opened 2023-05-16
V3. Added a toggle button to display the old SmartBench models in the factory settings

LettyYeti:


LettyYeti:
Also, please add a screengrab of the updated screen so we can check it visually :)

VladPavelescu:
### **Default view:**
image


### **Showing all SmartBench models:**
image

Tested locally and on rig by going to the factory screen and manually checking that the list of models changes correctly when pressing the toggle button. Also tested when the list of all models exceeds screen size, the Spinner automatically makes it into a scrollable list.

PR #1174, opened 2023-05-16
V4. Made popups in Factory screen auto dismiss

LettyYeti:


LettyYeti:
Update: Ed is happy with it, and I think it's ok too :)

PR #1175, opened 2023-05-16
V5. Changed checkbox icon to red to stand out

LettyYeti:


LettyYeti (commented):
I didn't realise that the checkbox icon was actually already different from whatever the kivy default is! So changing this image actually affects a lot of screens, and may look "wrong" because it's red, rather than a neutral colour. 

That said, I think it's still good if the general checkbox graphic is more visible. 

Can you do the same boldness but a dark blue? 

Sorry, I didn't realise how wide-ranging this change might be when I turned the feedback from customer support into a brief!

PR #1177, opened 2023-05-16
V7. Added pre-update warning message and some comments

LettyYeti:


LettyYeti:
This commit wasn't properly independent of the other commits in vlad_temp because of the comments on new changes from other commits. When I cherry-picked it it had a load of merge conflicts as well, and I think this is why.

I'm going to look over it anyway, and ignore any changes that I've reviewed in other PRs. However, prior to final review and merge it's definitely worth getting all the other PRs signed off, merged into master, and then master should be merged back into this branch so that we can be confident about what changes will be made in this PR :) (and avoid any unexpected conflicts/missed changes)

src/asmcnc/skavaUI/popup_info.py
```diff
@@ -514,310 +516,376 @@ def __init__(self, screen_manager, localization, description):
...
+
         popup.open()
-        
+
         Clock.schedule_once(reboot, 6)
-    
-class PopupSoftwareRepair(Widget):   
+
+# Popup asking the user if they are sure they want to continue with the update with a given warning_message
+class PopupSoftwareUpdateWarning(Widget):
```
LettyYeti, line 573: Because this is a brand new popup, and one that will need full translation checking, it would make sense to write a proper test script for visually checking it! This also makes development much easier, especially when designing more complex UIs, so it's good to get in the habit early :) 

[Example here](https://github.com/YetiTool/easycut-smartbench/blob/master/tests/manual_tests/visual_screen_tests/yetipilot_settings_popup_test.py)

LettyYeti (commented):
There's a few comments on here about processes you probably won't know about yet! Don't worry about that yet - work on the other PRs, and by the time you get to this one, I should have some docs together that makes the workflow a bit easier. I've just put some fairly quick comments on here as well, so if anything's unclear feel free to ask :). 

Because this is a new visual as well, this is exactly the kind of PR that needs a screengrab of the new visual :)

PR #1178, opened 2023-05-16
V8. Fixed infinite 'Please wait' popup when USB stick does not contain update file

LettyYeti:


LettyYeti (commented):
Might be worth changing the order on this so that the wait popup is dismissed after the next popup comes up (just in case of any slowness), but it's not a biggie :)

PR #1181, opened 2023-05-17
Add generic function for dollar setting

DennisKiselev:
Tested on windows

### DIFF
--- a/src/asmcnc/apps/systemTools_app/screens/screen_final_test.py
+++ b/src/asmcnc/apps/systemTools_app/screens/screen_final_test.py
@@ -279,8 +279,7 @@ class FinalTestScreen(Screen):
         try:
             x_overstep = float(self.x_over_count.text)*self.x_calibration_scale_factor
             print(x_overstep)
-            self.m.send_any_gcode_command("$100 = " + str(float(self.m.s.setting_100) - x_overstep))
-            self.m.send_any_gcode_command("$$")
+            self.m.write_dollar_setting(100, float(self.m.s.setting_100) - x_overstep)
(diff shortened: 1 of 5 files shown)


src/asmcnc/apps/systemTools_app/screens/screen_final_test.py
```diff
@@ -279,8 +279,7 @@ def set_x_steps(self):
         try:
             x_overstep = float(self.x_over_count.text)*self.x_calibration_scale_factor
             print(x_overstep)
-            self.m.send_any_gcode_command("$100 = " + str(float(self.m.s.setting_100) - x_overstep))
-            self.m.send_any_gcode_command("$$")
+            self.m.write_dollar_setting(100, float(self.m.s.setting_100) - x_overstep)
```
LettyYeti, line 282: Test these changes on the rig, just in case the reset after the stream causes any issues

LettyYeti (commented):
You've put "tested on windows" - is this with an arduino? Were you looking at output? More info please, as these changes obvs affect machine commands :)

src/asmcnc/apps/systemTools_app/screens/screen_final_test.py
DennisKiselev: Just to check, what kind of issues should I be looking for with the reset? It looks like the values are updating correctly on the rig so I assume that means there's no issue?

LettyYeti: Yep just checking for any mis-setting or weirdness :).

LettyYeti (approved):

PR #1187, opened 2023-05-25
Spindle test rig dennis

DennisKiselev:


src/asmcnc/comms/serial_connection.py
```diff
@@ -990,8 +990,9 @@ def process_grbl_push(self, message):
                         self.digital_spindle_mains_voltage = int(digital_spindle_feedback[3])
 
                         if self.sm.has_screen('spindle_test_1'):
-                            self.sm.get_screen('spindle_test_1').update_spindle_feedback()
-                            self.sm.get_screen('spindle_test_1').test.add_spindle_load()
+                            if self.digital_spindle_ld_qdA != -999:
```
LettyYeti, line 993: Is there any indicator for the user that this has happened? Do they need to know if the load is -999?

PR #1190, opened 2023-05-31
Generalise bake GRBL settings, and ZH cycling (for ZHQC and grbl settings screen)

LettyYeti:
Change to new ZH cage to accommodate double stack motors means that $132 = 130.0, instead of 150.0. 

Have set up GRBL settings bake to set the max Z travel based on the X motor current and FW version.

Have also set up error checking so that if X motor current does not get read in properly (and FW version is high enough that it should be), both ZHQC main (not warranty) and password protected grbl settings screen (in system tools) will show the operator a popup. 

ZHQC cycles (for main screen) have been generalised to use the grbl_z_max_travel value, which is read in from $132. 

Also changed current selection checkboxes so that double stack option is first and is checked by default. 

Tests: 
- Unit tests for get_z_max_travel_to_bake function

- GRBL bake from grbl settings screen tested on Dev machine; sort of held down stop bar at start up to prevent read in of registers to check popup failure condition. 

- Tested on ZHQC jig from FW flash through to calibration and the final ZH cycle. Also tested Bake GRBL if flashed single motor current (on double stack ZH) to go through both conditions.

### DIFF
--- a/src/asmcnc/production/z_head_qc_jig/z_head_qc_7.py
+++ b/src/asmcnc/production/z_head_qc_jig/z_head_qc_7.py
@@ -116,11 +116,11 @@ class ZHeadQC7(Screen):
         self.m.resume_from_alarm()
 
     def do_cycle(self):
-        self.m.s.write_command('G53 G0 Z-150')
+        self.m.s.write_command('G53 G0 Z-' + str(self.m.grbl_z_max_travel))
         self.m.s.write_command('G53 G0 Z-1')
-        self.m.s.write_command('G53 G0 Z-150')
+        self.m.s.write_command('G53 G0 Z-' + str(self.m.grbl_z_max_travel))
         self.m.s.write_command('G53 G0 Z-1')
-        self.m.s.write_command('G53 G0 Z-150')
+        self.m.s.write_command('G53 G0 Z-' + str(self.m.grbl_z_max_travel))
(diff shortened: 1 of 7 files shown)


src/asmcnc/production/z_head_qc_jig/z_head_qc_7.py
```diff
@@ -116,11 +116,11 @@ def resume_from_alarm(self):
...
 
     def do_cycle(self):
-        self.m.s.write_command('G53 G0 Z-150')
+        self.m.s.write_command('G53 G0 Z-' + str(self.m.grbl_z_max_travel))
         self.m.s.write_command('G53 G0 Z-1')
-        self.m.s.write_command('G53 G0 Z-150')
+        self.m.s.write_command('G53 G0 Z-' + str(self.m.grbl_z_max_travel))
         self.m.s.write_command('G53 G0 Z-1')
-        self.m.s.write_command('G53 G0 Z-150')
+        self.m.s.write_command('G53 G0 Z-' + str(self.m.grbl_z_max_travel))
```
archiepeterjarvis, line 123: This movement seems quite common, a function with parameter for Z could be valuable.

LettyYeti: @archiepeterjarvis I agree, but will make a new ticket for it :)

PR #1199, opened 2023-06-08
Rework spindle settings sc1

DennisKiselev:
Merging main layout for reworked spindle settings screen into master, with corrected tab image, shown below

image


LettyYeti:
image

src/asmcnc/apps/maintenance_app/widget_maintenance_spindle_save.py
```diff
@@ -82,30 +67,30 @@ def __init__(self, **kwargs):
...
                 "[b]: [/b]" + \
                 self.l.get_str("The spindle needs to cool down after a job to prevent it from overheating, and to extend its lifetime.") + \
                 " " + \
                 self.l.get_str("We recommend the following cooldown settings:") + \
-                "\n" + \
+                "\n\n" + \
                 "       " + \
-                self.l.get_str("Yeti: 20,000 RPM; 10 seconds") + \
-                "\n" + \
+                self.l.get_str("Yeti SC1/2: 12,000 RPM; 10 seconds") + \
```
LettyYeti, line 83: non blocking (bc it's an existing bug) but I've just realised that "RPM" and "seconds" should be replaced with their translated variants. Can you address?

PR #1204, opened 2023-06-10
Flatten spindle graphics

DennisKiselev:
Updated all recent graphics that have a spindle in them to be flat at the bottom, to keep it consistent across the software

Tested on windows

image
image
image
image
image
image
image


LettyYeti (dismissed):
Note that src/asmcnc/core_UI/job_go/img/health_check_button_big.png has a slightly darker background on the original background - I think this is important as I remember making a subtle change during the YP development to make sure it showed up well enough on the console :)

DennisKiselev:
image

PR #1205, opened 2023-06-13
Made network change without rebooting

VladPavelescu:
### This popup appears while the network is being changed in the background and goes away automatically once an IPv4 address has been allocated on the new network
image

I've tested it by switching between Yeti Wi-Fi and a personal hotspot and checking that the IP address has changed correctly and used `ping` to check Internet connectivity.

### DIFF
--- a/src/asmcnc/apps/wifi_app/screen_wifi.py
+++ b/src/asmcnc/apps/wifi_app/screen_wifi.py
@@ -457,9 +457,6 @@ class WifiScreen(Screen):
 
     def connect_wifi(self):
         self._password.text = ''
-
-        # Save the current IP address to dismiss the wait popup when the address changes
-        current_ip_address = self.ip_status_label.text
         wait_popup = popup_info.PopupWait(self.sm, self.l)
 
         # pass credentials to wpa_supplicant file
@@ -489,10 +486,12 @@ class WifiScreen(Screen):
             os.system('sudo systemctl restart dhcpcd')
 
             def dismiss_wait_popup(dt):
-                if self.ip_status_label.text != '' and self.ip_status_label.text != current_ip_address:
+                if self.set.wifi_available:
                     wait_popup.popup.dismiss()
+                    return
+                Clock.schedule_once(dismiss_wait_popup, 0.5)
 
-            Clock.schedule_interval(dismiss_wait_popup, 0.5)
+            Clock.schedule_once(dismiss_wait_popup, 5)
 
         except: 
             try: 
--- a/src/asmcnc/apps/wifi_app/screen_wifi.py
+++ b/src/asmcnc/apps/wifi_app/screen_wifi.py
@@ -614,6 +614,9 @@ class WifiScreen(Screen):
                 self.connection_instructions_rst.source = self.wifi_documentation_path + self.l.default_lang + '.rst'
 
     def on_leave(self):
-        self.wifi_error_timeout_event.cancel()
-        self.dismiss_wait_popup_event.cancel()
-        #self.refresh_ip_label_value_event.cancel()
+        if self.wifi_error_timeout_event:
+            self.wifi_error_timeout_event.cancel()
+        if self.dismiss_wait_popup_event:
+            self.dismiss_wait_popup_event.cancel()
+        if self.refresh_ip_label_value_event:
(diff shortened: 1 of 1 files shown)


src/asmcnc/apps/wifi_app/screen_wifi.py
```diff
@@ -476,6 +476,23 @@ def connect_wifi(self):
...
+            os.system('sudo wpa_cli -i wlan0 reconfigure')
+
+            # Restart the DHCP service to allocate a new IP address on the new network
+            os.system('sudo systemctl restart dhcpcd')
+
+            def dismiss_wait_popup(dt):
+                if self.set.wifi_available:
+                    wait_popup.popup.dismiss()
+                    return
+                Clock.schedule_once(dismiss_wait_popup, 0.5)
```
LettyYeti, line 492: This does need a timeout/fail condition (users will type in details wrong sometimes or possibly intentionally type in wrong creds to disconnect) 

"No WiFi connection!" has already been translated, so I recommend using that as the text on the error popup for now. 

It's worth adding some extra text for this error in the near future as well (along the lines of check your network name/password/try again/check your network/contact yetitool etc.), so I'd recommend writing the full error message and getting it translated asap. However, "No WiFi connection!" will pass for now, so that can be treated as a separate ticket on a new branch :)

src/asmcnc/apps/wifi_app/screen_wifi.py
```diff
@@ -503,7 +503,30 @@ def connect_wifi(self):
...
+        # Reload the updated wpa_supplicant file
+        os.system('sudo wpa_cli -i wlan0 reconfigure')
+
+        # Restart the DHCP service to allocate a new IP address on the new network
+        os.system('sudo systemctl restart dhcpcd')
+
+        def dismiss_wait_popup(dt):
+            if self.set.wifi_available:
+                wait_popup.popup.dismiss()
+                Clock.unschedule(wifi_error_timeout)
```
LettyYeti, line 518:
```
def my_callback(dt):
    pass

# call my_callback every 0.5 seconds
event = Clock.schedule_interval(my_callback, 0.5)

# call my_callback in 5 seconds
event2 = Clock.schedule_once(my_callback, 5)

event_trig = Clock.create_trigger(my_callback, 5)
event_trig()

# unschedule using cancel
event.cancel()

# unschedule using Clock.unschedule
Clock.unschedule(event2)

**# unschedule using Clock.unschedule with the callback
# NOT RECOMMENDED
Clock.unschedule(my_callback)**
```

from: https://kivy.org/doc/stable/api-kivy.clock.html

You can check other instances of unscheduling events in the code as well, but generally you want something along the lines of: 
event = Clock.schedule_.....
if self.event: Clock.unschedule(self.event)

so that you check that the event still exists when you unschedule it, otherwise it WILL cause a crash. 

You would probably also benefit from putting a try-except on any popup dismisses, because if the popup gets dismissed (i.e. by a user clicking off of it) before it's dismissed that will also probably cause a crash. 

Finally, it's good practice to make sure any events that may still be running are cleared up on_leave(). 

The wifi_error_timeout might even be better if it's called by the dismiss function after x amount of time has passed, because then that event never gets set up (and doesn't need to be cleared out) if the wifi setting is a success :)

LettyYeti (changes requested):
Also please reply to comments/ add more details about further testing for "unhappy path" scenarios :)

src/asmcnc/apps/wifi_app/screen_wifi.py
VladPavelescu: - I've now changed it so that the new block of code gets called after all try-except blocks.
- Due to the timeout, unless the connection is successful, the error popup will come up and warn the user that something went wrong.

src/asmcnc/apps/wifi_app/screen_wifi.py
VladPavelescu: All changes have been implemented apart from the last one as discussed. I have tested it on the rig with valid and invalid credentials, as well as disconnecting from network after pressing connect

LettyYeti (changes requested):
Try just opening and immediately exiting the WiFi app ;)

PR #1207, opened 2023-06-14
Wifi disconnect clear

VladPavelescu:
### The following popup appears when entering the Wi-Fi screen if there has previously been a working Wi-Fi connection.
image

Tested on the rig by connecting to the network, exiting the software, shutting down the wireless interface and restarting the software. Also tested after factory reset and it does not show the popup then.


src/asmcnc/apps/wifi_app/screen_wifi.py
```diff
@@ -503,6 +525,7 @@ def connect_wifi(self):
                 os.system('echo "update_config=1" | sudo tee --append /etc/wpa_supplicant/wpa_supplicant-wlan0.conf')
                 os.system('echo "country="' + self.country.text + '| sudo tee --append /etc/wpa_supplicant/wpa_supplicant-wlan0.conf')
 
+                os.system('sudo sed -i "s/wifi_connected_before=False/wifi_connected_before=True/" config.txt')
```
LettyYeti, line 528: Rather than repeating this line for each case, is it worth just having it once at the same place the network gets restarted (or after, if it's successful)?

src/asmcnc/apps/wifi_app/screen_wifi.py
```diff
@@ -455,6 +463,18 @@ def check_credentials(self):
         else: 
             self.connect_wifi()
 
+    def is_wlan0_connected(self):
+        #returns "state UP" or "state DOWN" depending on whether wlan0 is connected or not
+        state_raw = os.popen('ip addr show | grep "wlan0" | grep -oP "state\s\w+"').read()
+        state = state_raw.split(" ")[1].strip("\n")
+
+        if state == "UP":
```
LettyYeti, line 471: You can simplify this block to something like: 

print("wlan0 connection status: " + str(state))
return state == "UP"


>>> def func(arg): return arg==1
>>> func(0)
False
>>> func(1)
True

LettyYeti (changes requested):
See comments. 

Refactoring is minor but translations are important - we don't merge visible customer facing things until they've been translated :)

LettyYeti (approved):

PR #1215, opened 2023-06-21
Wifi ssid list

VladPavelescu:
Weird or duplicate network names should no longer appear in the list of networks in the Wi-Fi screen. Tested on rig by changing the name of the hotspot on my phone to include special characters like \n, \t, \x00, etc

### DIFF
--- a/src/asmcnc/apps/wifi_app/screen_wifi.py
+++ b/src/asmcnc/apps/wifi_app/screen_wifi.py
@@ -531,8 +531,8 @@ class WifiScreen(Screen):
         raw_SSID_list = os.popen('sudo iwlist wlan0 scan | grep "ESSID:" | sed "s/ESSID://g" | sed "s/^[ \t]*//g"').read()
         SSID_list = raw_SSID_list.replace('"','').strip().split('\n')  # Remove " from network name and split on newline
         if '' in SSID_list: SSID_list.remove('')  # Remove empty entries
-        SSID_list = [x for x in SSID_list if not set(x) <= set("\\x00")]  # Remove any addresses that contain only NULL bytes
-        SSID_list = set(SSID_list)  # Remove duplicate entries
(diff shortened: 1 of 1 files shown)


src/asmcnc/apps/wifi_app/screen_wifi.py
```diff
@@ -526,9 +526,13 @@ def quit_to_lobby(self):
...
     def get_available_networks(self):
-        raw_SSID_list = os.popen('sudo iw dev wlan0 scan | grep SSID').read()
-        SSID_list = raw_SSID_list.replace('\tSSID: ','').strip().split('\n')
-        if '' in SSID_list: SSID_list.remove('')
+        # Scan for networks, select only ESSIDs, remove ESSID from the line, remove any leading whitespaces or tabs.
+        # This leaves each network name in the format "NETWORK NAME" with each of them on their own new line
+        raw_SSID_list = os.popen('sudo iwlist wlan0 scan | grep "ESSID:" | sed "s/ESSID://g" | sed "s/^[ \t]*//g"').read()
+        SSID_list = raw_SSID_list.replace('"','').strip().split('\n')  # Remove " from network name and split on newline
+        if '' in SSID_list: SSID_list.remove('')  # Remove empty entries
+        SSID_list = [x for x in SSID_list if not set(x) <= set("\\x00")]  # Remove any addresses that contain only NULL bytes
```
archiepeterjarvis, line 534: You can make this a set therefore removing the duplicates by changing your list comprehension to set comprehension. Like this: 

`{x for x in SSID_list...}` 

Instead of using set() :)

LettyYeti: This sounds great, nice feedback @archiepeterjarvis 

Do you know whether it's likely to be faster than list comprehension as well? Might be worth running a timeit experiment just to see whether there are any small differences (partly so we know for future dev)

PR #1216, opened 2023-06-30
Refactor visual tests

DennisKiselev:
Refactored visual screen tests to run from one file, with generalised setup code and functions that represent individual tests and can be run from the command line

Tested on windows


LettyYeti (approved):
This is a great start, and it has already massively reduced the amount of duplication present in the previous "system" :). If you've run through all of the tests, then I'm happy for it to be merged. 

I think it can be taken further though, so that it's even more usable for future testing!

Next steps may include: 

- Generalising the Dummy Serial setup - this is already repeated a few times, and it might make sense to put it in a function with optional/default arguments (e.g. setting_50=0.03) that don't have to be specified, but can be set if the tester needs them for something specific 

- Generalising the BasicScreen class: at the moment it's a vessel for the YP settings popup. It would be ace if it could be re-used for any popup that needs to be tested

- Screen set up functions outside of test functions: so if you want to test a screen under a series of cases, you can call one function to create the screen and add it to the SM (rather than copy-pasting initialisation).

- Setting up a function that can create screens and add widgets - has some default arguments that can be modded (rather than typing out full syntax every time) - maybe it takes in just the class name as a main argument. 

- Possibly capability to run through multiple screens in a list that a tester can add to easily - e.g. for testing localisation on a number of screens

- Adding lots more comments, and putting screens into "groups" so that it's easy to scroll through the code and locate the test you need (or the test you need to write)

PR #1222, opened 2023-07-12
*DO NOT MERGE, PR FOR REVIEW COMMENTS* Localisation script

LettyYeti:


src/asmcnc/comms/localisation_script.py
```diff
@@ -0,0 +1,106 @@
...
+                    if foreign_line[default_lang].strip() == localisation_line[default_lang].strip():
+                        for language in supported_languages:
+                            row.append(localisation_line[language].strip())
+                        # writer.writerow doesn't deal well with quotes so writing directly to the file is easiest
+                        # writer.writerow(row)
+                        f.write('\t'.join(row) + '\n')
+                        localisation_lst.remove(localisation_line)
+                        break
+                else:
+                    failed_phrases.append(foreign_line[default_lang])
```
LettyYeti, line 68: Just to clarify, this is only called when the foreign line doesn't match to any localisation line, is that correct?

src/asmcnc/comms/localisation_script.py
```diff
@@ -0,0 +1,106 @@
...
+
+            # Get only the english phrases in the dictionary
+            english_lst = [x[default_lang] for x in foreign_lst]
+
+            try:
+                os.mkdir("logs")
+            except:
+                pass
+
+            with open("logs\\log " + datetime.now().strftime("%d-%m-%Y %H-%M-%S") + ".txt", "w",
```
LettyYeti, line 80: wrong slash here, which is why it's not putting it in the folder

LettyYeti (commented):
Just needs a fix for the log folder, and some questions that might benefit from being answered by some clarifying comments :) 

From the output it obvs seems to work well. 

I think when you've got a bit of time (maybe towards the end of the sprint), it would be worth writing a unit test file to cover this script so that it's fully maintainable and easy to modify going forwards :). I expect that that will lead to a bit of a re-factor that puts each step into a function as well. 

Don't merge into master, as per our discussions about popping it in a separate repo for benefit of us devs :)

src/asmcnc/comms/localisation_script.py
```diff
@@ -0,0 +1,106 @@
...
+from datetime import datetime
+
+if sys.version_info < (3, 1):
+    print("This script requires Python version 3.1 or higher")
+    exit(0)
+
+"""
+This script requires Python 3.1 or higher. Edit the variables below to change the settings on the script. 
+Download the .tsv file from the Localisation Master spreadsheet and update the path variables below.
+Once all the variables are updated correctly, simply run the script. It will generate a log file (in a logs folder) to
```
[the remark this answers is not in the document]
VladPavelescu: Log files are getting saved to a `logs` folder on Windows with expected names... I will look into it

src/asmcnc/comms/localisation_script.py
```diff
@@ -0,0 +1,106 @@
...
+new_strings = False
+
+# File paths of the current foreign dictionary and the localisation master
+foreign_dictionary = "foreign_dictionary.txt"
+localisation_master = "Localisation Master - Human edits.tsv"
+
+# File path of the new dictionary file to write
+new_file = "new.txt"
+
+default_lang = 'English (GB)'
```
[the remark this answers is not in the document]
VladPavelescu: I'm not sure if the script will work when this is modified (I think yes?) but I mainly created the variable for ease of writing code

src/asmcnc/comms/localisation_script.py
VladPavelescu: Yep. For each line in the foreign dictionary, it will try to find a match in the whole localisation master. If it doesn't find a match in the localisation master, it marks it as a fail so we can find missing phrases easier

PR #1237, opened 2023-09-06
File unit interpretation

DennisKiselev:
Two different fixes related to G20/G21 unit gcodes:

- When a stream ends, a G21 is sent to change the machine back to mm, in case the file used inches
- When using the GO XY button in job recovery, the last G20/G21 command is accounted for so that the position is accurately found.

Tested on windows, by using manual move and doing 1mm movements and seeing if it goes way too far

### DIFF
--- a/src/asmcnc/comms/serial_connection.py
+++ b/src/asmcnc/comms/serial_connection.py
@@ -639,6 +639,9 @@ class SerialConnection(object):
 
         log("Ending stream...")
 
+        # Always switch to mm, in case the job put you in inches
+        self.write_command('G21')
+
         # Reset flags
         self.is_job_streaming = False
         self.is_stream_lines_remaining = False
--- a/src/asmcnc/skavaUI/screen_job_recovery.py
+++ b/src/asmcnc/skavaUI/screen_job_recovery.py
@@ -330,6 +330,8 @@ class JobRecoveryScreen(Screen):
     scroll_up_event = None
     scroll_down_event = None
 
+    using_inches = False
+
     def __init__(self, **kwargs):
         super(JobRecoveryScreen, self).__init__(**kwargs)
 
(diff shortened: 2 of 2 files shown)


src/asmcnc/comms/serial_connection.py
```diff
@@ -639,6 +639,9 @@ def end_stream(self):
 
         log("Ending stream...")
 
+        # Always switch to mm, in case the job put you in inches
+        self.write_command('G21')
```
LettyYeti, line 643: Would be good to have this (for both unit options) as a readable router_machine function, like zUp, spindle_off etc.

src/asmcnc/comms/serial_connection.py
```diff
@@ -639,6 +639,9 @@ def end_stream(self):
...
+
```
LettyYeti, line 644: Also need to add unit change to cancel_stream function as well as end_stream

LettyYeti (changes requested):

src/asmcnc/skavaUI/screen_job_recovery.py
```diff
@@ -584,13 +594,18 @@ def get_info(self):
...
             -self.m.limit_switch_safety_distance,
         )
+        # If Z is below safe height, then raise it up
         if self.m.mpos_z() < z_safe_height:
             self.m.s.write_command("G53 G0 Z%s F750" % z_safe_height)
-        self.m.s.write_command("G90 G0 X%s Y%s" % (self.pos_x, self.pos_y))
+        if self.using_inches:
+            self.m.set_machine_unit_to_inch()
+        self.m.s.write_command('G90 G0 X%s Y%s' % (self.pos_x, self.pos_y))
+        self.m.set_machine_unit_to_mm()
```
Masc999, line 608: Why do you switch back to mm if the recovered file is in inches?  Doesn't that mean, that when you re-run the job from the recover position all subsequent moves will be in mm again?

DennisKiselev: This function is just used for the GO XY button on the job recovery screen - it sets to inches just while it moves to the position currently being displayed on that screen, then it needs to go back to mm, and when the user runs the job recovery file, then it will insert the G20 back in

PR #1242, opened 2023-09-13
Bootloader upgrade

archiepeterjarvis:
This is the fix for the text colour bug on Pi4.

Tested on a fresh platform on my rig.

image

### DIFF
--- a/ansible/init.yaml
+++ b/ansible/init.yaml
@@ -157,19 +157,6 @@
       name: proftpd.service
       state: restarted
     ignore_errors: yes
-  - name: Upgrade raspberrypi bootloader
-    become: yes
-    become_user: pi
-    tasks:
-      - name: Update apt cache
-        apt:
-          update_cache: yes
-
-      - name: Upgrade raspberrypi-bootloader
-        apt:
-          name: raspberrypi-bootloader
-          state: latest
-    when: ansible_distribution_major_version == '11'
   - name: Generate /home/pi/.ssh/ RSA host key
     command : ssh-keygen -q -t rsa -b 4096 -f /home/pi/.ssh/id_rsa -N ""
     become: true
@@ -206,3 +193,16 @@
       state: present
       insertafter: EOF
       line: 'ansible_from_easycut=True'
+  - name: upgrade bootloader
+    become: yes
+    become_user: pi
+    tasks:
+      - name: Update apt cache
+        apt:
+          update_cache: yes
+
+      - name: Upgrade raspberrypi-bootloader
+        apt:
+          name: raspberrypi-bootloader
+          state: latest
+    when: ansible_distribution_major_version == '11'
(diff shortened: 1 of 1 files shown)


LettyYeti:
Ok. So. Bootloader update. i.e., fixes for grey flashy text and rainbow splash screen.
I was testing for the fail condition (i.e. it not installing), and then I installed the bootloader... and then I rebooted... and then it was giving no signs of life after a good long while, so I pulled the plug (bad, I know, but it was AGES) and now it's bricked and the RPi is continuously flashing green which means:

> If the green light constantly flashes in a regular pattern, it will typically indicate that the bootloader cannot detect an SD Card with a valid “start.elf” file.

Elliot also mentioned that updating apt breaks things for the big consoles: 

> Screen Notes: Apt Update issues (Fix text)
If you want to fix text and run something like the below
sudo apt-get update		or		sudo apt update
sudo apt-get full-upgrade	or		sudo apt upgrade
> 
> Some files of the originally installed driver will be overwritten, in this case, the display driver will need to be reinstalled to display normally (e.g> Use parts of option 2 of how to replace a Micro SD card)

So further work is needed on understanding the fail conditions around this before we can merge.  
- do we need a specific expectation around how long it will take to reboot after installing the update?
- what happens for big consoles? what workarounds can we use?

LettyYeti (changes requested):
Ok. So. Bootloader update. i.e., fixes for grey flashy text and rainbow splash screen.
I was testing for the fail condition (i.e. it not installing), and then I installed the bootloader... and then I rebooted... and then it was giving no signs of life after a good long while, so I pulled the plug (bad, I know, but it was AGES) and now it's bricked and the RPi is continuously flashing green which means:

> If the green light constantly flashes in a regular pattern, it will typically indicate that the bootloader cannot detect an SD Card with a valid “start.elf” file.

Elliot also mentioned that updating apt breaks things for the big consoles: 

> Screen Notes: Apt Update issues (Fix text)
If you want to fix text and run something like the below
sudo apt-get update		or		sudo apt update
sudo apt-get full-upgrade	or		sudo apt upgrade
> 
> Some files of the originally installed driver will be overwritten, in this case, the display driver will need to be reinstalled to display normally (e.g> Use parts of option 2 of how to replace a Micro SD card)

So further work is needed on understanding the fail conditions around this before we can merge.  
- do we need a specific expectation around how long it will take to reboot after installing the update?
- what happens for big consoles? what workarounds can we use?

LettyYeti:
It is most likely the display drivers, not the pi failing, but we still need a workaround for console XL before merge https://www.waveshare.com/wiki/10.1inch_DSI_LCD_(C)

LettyYeti:
Currently can't re-install the drivers for waveshare screen after the update, so we're waiting on waveshare for further news. This PR is on hold until we have a strategy that is not going to "brick" any consoles :) 

https://forums.raspberrypi.com/viewtopic.php?t=355034

LettyYeti (approved):

PR #1254, opened 2023-09-27
Wifi confirmation popup

VladPavelescu:
This popup comes up after the console has successfully connected to WiFi. Tested on rig, with all currently supported languages
image


LettyYeti (changes requested):
Can you re-generate the dictionary please, and ask Gemma to request Korean translation? Otherwise looks ok

PR #1260, opened 2023-10-02
Translate 'Korean' into Korean, and slight refactor for how language names are checked

LettyYeti:
Tested on PC and on rig, both with visual screen tests & in live software. 

Note that Korean characters automatically worked in visual screen tests, but not in actual SW for welcome screen.


src/asmcnc/apps/start_up_sequence/welcome_to_smartbench_app/screens/screen_language_select.py
```diff
@@ -379,7 +379,11 @@ def change_welcome_label(self, dt):
 		if self.welcome_i < 6:
 			self.welcome_i += 1
 
+		elif self.welcome_i == 7:
```
DennisKiselev, line 382: How is this condition triggered? Wouldn't it go to the else condition when `self.welcome_i == 6` and get set back to 0?

src/asmcnc/apps/start_up_sequence/welcome_to_smartbench_app/screens/screen_language_select.py
```diff
@@ -411,6 +415,10 @@ def next_screen(self):
 		self.start_seq.next_in_sequence()
 
 	def update_strings(self):
+		if self.l.lang == self.l.ko:
```
DennisKiselev, line 418: The font is not set back to default (not a blocking issue though)

LettyYeti: fixed

PR #1269, opened 2023-10-11
Keyboard margin increase

VladPavelescu:
Increased the margin on top of the keyboard when it is allowed to be dragged around so that it's easier to grab and move. Also added background images for the keyboard to make it obvious to users that they can move the keyboard.

image

image


src/asmcnc/keyboard/custom_keyboard.py
```diff
@@ -110,14 +113,44 @@ def key_up(self, keycode, internal, modifiers):
...
+                        self.pos = (self.pos[0], 0)
+                    if self.pos[0] < 0:
+                        self.pos = (0, self.pos[1])
+                    if self.pos[1] + self.height > Window.height:
+                        self.pos = (self.pos[0], Window.height - self.height)
+                return
             self.text_instance.insert_text(internal)
 
+    def set_keyboard_background(self):
+        if self.do_translation == (True,True):
```
LettyYeti, line 145: why is this a tuple, when further up it's just set to 'True'?

LettyYeti (changes requested):
- Can the numpad keyboard layout be modified so that it's 123 across the top, 0 in the bottom middle (like phone/card machine numpads)
- Does the numpad need an enter button?

src/asmcnc/keyboard/custom_keyboard.py
VladPavelescu: - I will change the layout of the numpad, I didn't realise it's quite stupid at the moment.
- Since the user will never get the numpad on its own for the moment, I don't think so. I've also talked to Ed about it and he said he doesn't think there's any instance where there isn't an "enter" button on the screen for any text inputs. Adding one is really easy though, functionality is the same as normal enter. Would you like me to add one?
- I don't know why kivy treats it as a tuple either and it took a little while to realise that it's not the same as when it's set. I assume `do_translation` is stored as a tuple because it has a boolean for vertical and horizontal translation.

VladPavelescu: Is this okay? 

image

LettyYeti (approved):

PR #1272, opened 2023-10-19
Automatically set screen resolution

VladPavelescu:
Automatically detects the resolution of the screen on the raspberry pi's and defaults to 800x480 for all other operating systems.

Tested on laptop and console

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -28,7 +28,7 @@ from kivy.clock import Clock
 Config.set('kivy', 'keyboard_mode', 'systemanddock')
 
 if sys.platform.startswith("linux"):
-    # get screen resolution as "1280x800"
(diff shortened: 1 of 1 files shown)


src/main.py
```diff
@@ -26,8 +26,17 @@
 from kivy.config import Config
 from kivy.clock import Clock
 Config.set('kivy', 'keyboard_mode', 'systemanddock')
-Config.set('graphics', 'width', '800')
-Config.set('graphics', 'height', '480')
+
+if sys.platform.startswith("linux"):
+    # get screen resolution as "1280x800"
```
LettyYeti, line 31: or 800x480

LettyYeti (changes requested):

LettyYeti (approved):

PR #1275, opened 2023-11-01
Drywall app cfg - DRAFT PR

LettyYeti:


src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
```diff
@@ -0,0 +1,104 @@
...
+        configurations_dir = 'src/asmcnc/apps/drywall_cutter_app/config/configurations'
+        cutters_dir = 'src/asmcnc/apps/drywall_cutter_app/config/cutters'
+
+
+class DWTConfig(object):
+    active_config = None  # type: config_classes.Configuration
+    active_cutter = None  # type: config_classes.Cutter
+
+    def __init__(self):
+        fix_paths_for_tests()
```
LettyYeti, line 21: Can this be called from the test files? It seems like this risks causing havoc later on.

src/asmcnc/apps/drywall_cutter_app/screen_drywall_cutter.py
```diff
@@ -99,29 +100,31 @@
...
     line_cut_options = ['Cut on line', 'Cut inside line', 'Cut outside line']
+    dwt_config = config_loader.DWTConfig()
 
     def __init__(self, **kwargs):
         super(DrywallCutterScreen, self).__init__(**kwargs)
 
         self.sm = kwargs['screen_manager']
         self.m = kwargs['machine']
-        self.l = kwargs['localization']
+        # self.l = kwargs['localization']
```
LettyYeti, line 120: Why has this been commented out?

src/asmcnc/apps/drywall_cutter_app/screen_drywall_cutter.py
```diff
@@ -146,3 +149,6 @@ def save(self):
 
     def run(self):
         pass
+
+    def on_leave(self, *args):
```
LettyYeti, line 153: Is a matching on_enter needed to load the temp config?

tests/manual_tests/visual_screen_tests/test_dwt_parameter_change_hook.py
```diff
@@ -0,0 +1,58 @@
...
+from kivy.lang import Builder
+from kivy.uix.screenmanager import ScreenManager, Screen, NoTransition
+
+try:
+    import unittest
+    import pytest
+    from mock import Mock, MagicMock
+except:
+    print("Can't import mocking packages, are you on a dev machine?")
+
```
LettyYeti, line 16: Would be good to have a comment on how to run

LettyYeti (commented):
This looks great! 

Only other thing apart from the in line comments is that if I try to run the tests I get "Exception: Cutter file does not exist" :)

src/asmcnc/apps/drywall_cutter_app/screen_drywall_cutter.py
archiepeterjarvis: No, on init the dwt config module will load it if it exists.

src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
archiepeterjarvis: Good idea, thanks

archiepeterjarvis: Didn't work when I called it from the test instead :/

archiepeterjarvis:
> This looks great!
> 
> Only other thing apart from the in line comments is that if I try to run the tests I get "Exception: Cutter file does not exist" :)

src/asmcnc/apps/drywall_cutter_app/screen_drywall_cutter.py
LettyYeti: ah but what if the user goes in and out of the app? surely init will only run at start up?

src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
LettyYeti: Ok, then do a check to see if the paths exist before setting up the test case? Then don't need to comment anything out, and should be able to interchange between tests and actual use cases :)

src/asmcnc/apps/drywall_cutter_app/screen_drywall_cutter.py
archiepeterjarvis: If the screen is disposed, when the screen is created __init__ will be run again. If the screen isn't disposed, the object will still be there.

src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
archiepeterjarvis: Done, it now changes the paths in the test file

PR #1289, opened 2023-11-15
Check config files are valid before loading them

archiepeterjarvis:
Tested via unit tests

### DIFF
--- a/src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
+++ b/src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
@@ -9,7 +9,7 @@ cutters_dir = 'asmcnc/apps/drywall_cutter_app/config/cutters'
 DEBUG_MODE = True
 
 
-def debug_decorator(func):
+def debug(func):
     def wrapper(*args, **kwargs):
         if DEBUG_MODE:
             print('Calling function: ' + func.__name__ + ' with args: ' + str(args) + ' and kwargs: ' + str(
@@ -32,7 +32,7 @@ class DWTConfig(object):
             self.active_cutter = self.load_cutter(self.active_config.cutter_type)
 
     @staticmethod
-    @debug_decorator
+    @debug
     def is_valid_configuration(config_name):
         # type (str) -> bool
         """
@@ -60,7 +60,7 @@ class DWTConfig(object):
         return True
 
     @staticmethod
-    @debug_decorator
+    @debug
     def get_missing_fields(config_name):
         # type (str) -> list[str]
         """
@@ -89,7 +89,7 @@ class DWTConfig(object):
         return missing_fields
 
     @staticmethod
-    @debug_decorator
+    @debug
     def fix_config(config_name):
         # type (str) -> bool
         """
@@ -118,7 +118,7 @@ class DWTConfig(object):
 
         return True
 
-    @debug_decorator
+    @debug
     def load_config(self, config_name):
         # type (str) -> None
         """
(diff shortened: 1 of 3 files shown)


src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
```diff
@@ -11,8 +12,10 @@
 def debug_decorator(func):
     def wrapper(*args, **kwargs):
         if DEBUG_MODE:
-            print('Calling function: ' + func.__name__ + ' with args: ' + str(args[-1:]) + ' and kwargs: ' + str(kwargs))
```
archiepeterjarvis, line 14: When calling static function the [-1:] was removing first arg (which wasn't self in this case).

src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
```diff
@@ -28,7 +31,94 @@ def __init__(self):
...
+        valid_field_count = len(inspect.getargspec(config_classes.Configuration.__init__).args) - 1
+
+        if field_count != valid_field_count:
+            return False
+
+        return True
+
+    @staticmethod
+    @debug
+    def get_missing_fields(config_name):
```
LettyYeti, line 64: This doesn't look like it's used or tested anywhere?

tests/automated_unit_tests/apps/dwt/test_dwt_config_checking.py
```diff
@@ -0,0 +1,93 @@
...
+        dwt_config = config_loader.DWTConfig()
+
+        self.assertFalse(dwt_config.is_valid_configuration(self.invalid_temp_json_file.name))
+
+    def test_fix_valid_config(self):
+        dwt_config = config_loader.DWTConfig()
+
+        dwt_config.fix_config(self.invalid_temp_json_file.name)
+
+        self.assertTrue(dwt_config.is_valid_configuration(self.valid_temp_json_file.name))
```
LettyYeti, line 89: The filename here is different to the one used for fix_config - this looks wrong?

src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
```diff
@@ -28,7 +31,94 @@ def __init__(self):
...
+
+        if not os.path.exists(file_path):
+            return False
+
+        with open(file_path, 'r') as f:
+            cfg = json.load(f)
+
+        valid_field_names = inspect.getargspec(config_classes.Configuration.__init__).args[1:]
+
+        for field_name in valid_field_names:
```
LettyYeti, line 112: Because of the for loop here, please test (or ask someone to test) on a Pi, with a fully failing file, as the blocking for loop could cause lag on loading.

LettyYeti (changes requested):

tests/automated_unit_tests/apps/dwt/test_dwt_config_checking.py
archiepeterjarvis: Good spot, thanks

src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
archiepeterjarvis: Have tested on console, loads pretty much instantly even with an empty file.

tests/automated_unit_tests/apps/dwt/test_dwt_config_checking.py
LettyYeti: @archiepeterjarvis shouldn't it be corrected to self.invalid_temp_json_file.name for the fix_config to do something?

LettyYeti (approved):

PR #1293, opened 2023-11-17
Hide temp_config.json from filechooser

archiepeterjarvis:
I tried using the 'filter' property on the filechooser but 1.11.0 doesn't support a func filter being assigned.

Tested on console & via unit tests


src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
```diff
@@ -6,6 +6,7 @@
 configurations_dir = 'asmcnc/apps/drywall_cutter_app/config/configurations'
 cutters_dir = 'asmcnc/apps/drywall_cutter_app/config/cutters'
 
+TEMP_CONFIG_PATH = os.path.join(configurations_dir, '..', 'temp', 'temp_config.json')
```
LettyYeti, line 9: Maybe I am missing something, but this looks different to what's in

` self.save_config(os.path.join('..', 'temp', 'temp_config.json'))`

and 

`self.load_config(os.path.join('..', 'temp', 'temp_config.json'))`

archiepeterjarvis: That's because load_config joins the config directory and the argument passed.

LettyYeti (approved):

PR #1295, opened 2023-11-18
Material setup popup

VladPavelescu:
## **!!!! IGNORE THIS - OLD PR**
Material setup popup. Tested on laptop and rig. There are still a few test cases which I'm waiting to get an answer for but the popup is mainly finished.

image


archiepeterjarvis (commented):
There are a lot of try excepts in here, all with 'pass' which can make future debugging difficult.

Is it possible to change these try's into if statements that check for your exception cause? (if that makes sense)

LettyYeti (commented):
See doc: https://docs.google.com/document/d/1zU_bOkyl8UNqZDof8Zo4yBWCRBoBYLiGflymRFsfoAc/

src/asmcnc/keyboard/custom_keyboard.py
```diff
@@ -152,13 +136,33 @@ def set_keyboard_background(self):
...
+
     # On focus behaviour is bound to all text inputs
     def on_focus_raise_keyboard(self,instance,value):
         if value:
             try:
                 instance.get_focus_previous().focus = False
             except:
                 pass
+            # input_filter can be either None, ‘int’ (string), or ‘float’ (string), or a callable.
+            if (instance.input_filter != None):
```
archiepeterjarvis, line 162: redundant parenthesis & comparison
`if instance.input_filter`

PR #1285, opened 2023-11-08
Drywall canvas merge branch

DennisKiselev:
Canvas widget to display user's shape selections, and takes text inputs for dimensions, all connected with JSON backend

Tested on windows

PR #1296, opened 2023-11-18
Drywall canvas further changes

DennisKiselev:
**This one needs drywall_canvas_merge_branch to be merged first, as well as Vlad and Benji's branches**#
https://github.com/YetiTool/easycut-smartbench/pull/1285 (unsure where Vlad's and Benji's stuff is)

Added the go to datum and set datum buttons to drywall app XY move widget. Added a new function to router machine to perform the go to datum movement. Also implemented the flashing overlay for when the machine is not at the datum position.

Tested on windows


src/asmcnc/apps/drywall_cutter_app/widget_xy_move_drywall.py
```diff
@@ -179,9 +229,12 @@ def __init__(self, **kwargs):
         super(XYMoveDrywall, self).__init__(**kwargs)
         self.m=kwargs['machine']
         self.sm=kwargs['screen_manager']
+        self.l=kwargs['localization']
 
         self.set_jog_speeds()
 
+        Clock.schedule_interval(self.check_zh_at_datum, 0.04)
```
archiepeterjarvis, line 236: Avoid clock usage like this. Use event based approach that we discussed in sprint meet.

archiepeterjarvis (dismissed):

PR #1298, opened 2023-11-18
Drywall canvas square conversion

DennisKiselev:
Automatically converts rectangles to squares when both length inputs are equal

Tested on windows

### DIFF
--- a/src/asmcnc/apps/drywall_cutter_app/widget_drywall_shape_display.py
+++ b/src/asmcnc/apps/drywall_cutter_app/widget_drywall_shape_display.py
@@ -280,14 +280,20 @@ class DrywallShapeDisplay(Widget):
         self.dwt_config.on_parameter_change('canvas_shape_dims.r', float(value or 0))
 
     def x_input_change(self, instance, value):
-        if self.rotation_required():
-            self.sm.get_screen('drywall_cutter').rotate_shape(swap_lengths=False)
+        self.do_rectangle_checks()
         self.dwt_config.on_parameter_change('canvas_shape_dims.x', float(value or 0))
 
     def y_input_change(self, instance, value):
+        self.do_rectangle_checks()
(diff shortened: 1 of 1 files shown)


src/asmcnc/apps/drywall_cutter_app/widget_drywall_shape_display.py
```diff
@@ -280,15 +284,22 @@ def r_input_change(self, instance, value):
...
         self.dwt_config.on_parameter_change('canvas_shape_dims.y', float(value or 0))
 
+    def do_rectangle_checks(self):
+        if not self.swapping_lengths:
+            if self.rotation_required():
+                self.sm.get_screen('drywall_cutter').rotate_shape(swap_lengths=False)
+            if self.rectangle_with_equal_sides():
+                toolpath = self.sm.get_screen('drywall_cutter').cut_offset_selection.text
+                self.sm.get_screen('drywall_cutter').shape_selection.text = 'square'
+                self.sm.get_screen('drywall_cutter').cut_offset_selection.text = toolpath
```
LettyYeti, line 301: - I assume that it automatically detects when the shape_selection changes to 'square', and updates the canvas image accordingly? 

- Minor note that this function calls two external functions that each check the shape is a rectangle (and these functions aren't called anywhere else) before doing their main operation.

LettyYeti (commented):
Minor comments

src/asmcnc/apps/drywall_cutter_app/widget_drywall_shape_display.py
DennisKiselev: - Changing the text of a dropdown calls the on_text function like if you were to select something in the UI
- I think it makes sense logically to check the shape inside of the individual functions, for example if more shapes are added and the rotation_required function needs to be used for that too

PR #1402, opened 2023-12-15
C10 la calibration app screen distance 4 y

LettyYeti:
Tested on C7 & C10


src/asmcnc/calibration_app/screen_distance_4_y.py
```diff
@@ -298,6 +303,7 @@ def quit_calibration(self):
         self.sm.current = "tape_measure_alert"
 
     def next_screen(self):
+        # set up distance screen 1-x to return to after homing
```
BenjiPestell, line 306: is 1-x correct in this comment? following lines refer to 1_y

PR #1431, opened 2023-12-19
C10 bp popup data wifi warning

BenjiPestell:


BenjiPestell:
Tested on console 7 + 10

PR #1577, opened 2024-02-29
Benji/path utils

BenjiPestell:
Making relative paths a little easier :)

### DIFF
--- a/src/asmcnc/core_UI/path_utils.py
+++ b/src/asmcnc/core_UI/path_utils.py
@@ -1,0 +1,42 @@
+import os
+
+def get_path(target_dir):
+    """
+    Returns the path to the target directory.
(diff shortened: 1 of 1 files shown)


src/asmcnc/core_UI/path_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        target_dir (str): The target directory.
+
+    Returns:
+        str: The path to the target directory.
+    """
+    
+    root_path = os.path.abspath(__file__)  # Get the absolute path of the current script
+
+    try:
+        target_path_index = root_path.index(target_dir)
```
archiepeterjarvis, line 43: What if I need to get a folder that isn't on this files path?

PR #1585, opened 2024-03-06
Benji/path utils update

BenjiPestell:
Fix path Utils

### DIFF
--- a/src/asmcnc/core_UI/test_path_utils.py
+++ b/src/asmcnc/core_UI/test_path_utils.py
@@ -3,29 +3,23 @@ from path_utils import get_path
 
 class TestPathUtils(unittest.TestCase):
 
-    def setUp(self):
-        self.root = "/c:/Users/benji/Documents/"
-
     def test_get_path_files_only(self):
         target_dir = "router_machine.py"
-        result = get_path(target_dir, files_only=True)
-        self.assertEqual(result, self.root + "/easycut-smartbench/src/asmcnc/comms/router_machine.py")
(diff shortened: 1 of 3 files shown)


src/asmcnc/core_UI/test_path_utils.py
```diff
@@ -0,0 +1,36 @@
+import unittest
+from path_utils import get_path
+
+class TestPathUtils(unittest.TestCase):
```
archiepeterjarvis, line 4: beautiful !!

archiepeterjarvis: but please move to tests directory ;)

BenjiPestell: Done :)

tests/test_path_utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+sys.path.insert(0, '../src/asmcnc/core_UI')
+from path_utils import get_path
+
+class TestPathUtils(unittest.TestCase):
+
+    def test_get_path_files_only(self):
+        target_dir = "router_machine.py"
+        expected_result = "/easycut-smartbench/src/asmcnc/comms/router_machine.py"
+        result = get_path(target_dir, files_only=True).replace("\\", "/")
```
Masc999, line 13: the replace should not be needed. get_path and search_tree both take care.

BenjiPestell: Fixed :)

PR #1617, opened 2024-03-19
Archie/sb values fix

archiepeterjarvis:
Change paths to use path_utils and join instead of just concatenating. Stops multiple sb_values folders from appearing when running from different cwds. Searched repository for folders using find . -type d -name 'sb_values' and only got one after running from different wds.

### DIFF
--- a/src/asmcnc/comms/router_machine.py
+++ b/src/asmcnc/comms/router_machine.py
@@ -9,6 +9,7 @@ import traceback
 
 from enum import Enum
 from asmcnc.comms.logging_system.logging_system import Logger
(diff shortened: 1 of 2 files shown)


Masc999 (changes requested):
We now have to make sure that the sb_values folder exists. A fresh clone will crash at startup because GRBLSettingsManager and ModelManage access the path before router_machine is able to create it.
With this change even router_machine will crash when trying to access path_utils.sb_values_path

archiepeterjarvis:
Closing until solution is found :)

PR #1629, opened 2024-03-25
Benji/home minirig fix

BenjiPestell:
Fix homing for MiniRig (and other windows-connected machines)

### DIFF
--- a/src/asmcnc/skavaUI/screen_homing_active.py
+++ b/src/asmcnc/skavaUI/screen_homing_active.py
@@ -108,7 +108,10 @@ class HomingScreenActive(Screen):
 
     def on_enter(self):
         if sys.platform == "win32" or sys.platform == "darwin":
-            return
+            if self.m.m_state != "unknown": # Machine actually connected
+                self.m.do_standard_homing_sequence()
+            else: # No machine connected
+                return
         if self.m.homing_interrupted:
             return
         if not self.m.homing_in_progress:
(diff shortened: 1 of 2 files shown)


src/asmcnc/skavaUI/screen_homing_active.py
```diff
@@ -108,7 +108,10 @@ def on_pre_enter(self):
 
     def on_enter(self):
         if sys.platform == "win32" or sys.platform == "darwin":
-            return
+            if self.m.s.is_connected:
```
archiepeterjarvis, line 111: Do you need to check is_connected after checking platform? I assume if not connected, this function should do nothing. So it could be..: 

if not self.m.s.is_connected:
    return

....

archiepeterjarvis: But if you want to avoid additional testing, ignore me :)

src/asmcnc/skavaUI/screen_homing_active.py
```diff
@@ -170,5 +168,6 @@ def update_strings(self):
     def windows_cheat_to_procede(self):
         if sys.platform == "win32" or sys.platform == "darwin":
             self.return_to_ec_if_homing_not_in_progress()
+            self.m.is_machine_homed = True
         else:
             pass
```
archiepeterjarvis, line 173: Remove the else: pass also while here ;)

BenjiPestell: Hahaha

PR #1650, opened 2024-03-30
Archie/dwt/popup refactor

archiepeterjarvis:
Beautify/refactor the DWT validation popup.

image


src/asmcnc/comms/localization.py
```diff
@@ -22,16 +23,17 @@
...
-
 builder_font_string = """
 <Widget>:
     font_name: "%s"
     title_font: "%s"
 """
 
 
-
 class Localization(object):
```
Masc999, line 33: Please add a doc string for the class. Especially as it is a singlteton now.

src/asmcnc/core_UI/new_popups/job_validation_popup.py
```diff
@@ -0,0 +1,38 @@
...
+from kivy.uix.boxlayout import BoxLayout
+from kivy.uix.button import Button
+from kivy.uix.label import Label
+
+from asmcnc.comms.localization import Localization
+from asmcnc.core_UI import scaling_utils
+from asmcnc.core_UI.new_popups.popup_bases import PopupBase, PopupErrorTitle, PopupScrollableBody
+
+
+class JobValidationPopup(PopupBase):
```
Masc999, line 11: Add docstring and decribe what this is used for and how to use it.

src/asmcnc/core_UI/new_popups/job_validation_popup.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self, text, **kwargs):
+        super(JobValidationPopup, self).__init__(**kwargs)
+        localisation = Localization()
+
+        title = PopupErrorTitle(size_hint_y=0.15, localisation=localisation)
```
Masc999, line 16: Why not use localization as singleton directly in the class and not forwarding it?

src/asmcnc/comms/localization.py
```diff
@@ -74,14 +76,23 @@ class Localization(object):
     ORIGINAL_PRODUCT_NAME = "SmartBench"
     PRODUCT_NAME = "SmartBench"
 
+    def __new__(cls):
```
Masc999, line 79: 👍

src/asmcnc/core_UI/new_popups/job_validation_popup.py
archiepeterjarvis: It is always instantiated from a class which will have localisation. Acquiring the lock is an expensive operation so I thought this makes sense.

archiepeterjarvis: But if you think otherwise happy to change it 😁

Masc999: 👍

PR #1674, opened 2024-04-04
Benji/dwt/fix simulation

BenjiPestell:
# Fix simulation stop bug

### DIFF
--- a/src/asmcnc/apps/drywall_cutter_app/screen_drywall_cutter.py
+++ b/src/asmcnc/apps/drywall_cutter_app/screen_drywall_cutter.py
@@ -384,7 +384,7 @@ class DrywallCutterScreen(Screen):
         self.sm.current = 'lobby'
 
     def simulate(self):
-        self.popup_watchdog = Clock.schedule_interval(self.set_simulation_popup_state(self.m.s.m_state), 1)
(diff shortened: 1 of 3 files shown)


src/asmcnc/apps/drywall_cutter_app/screen_drywall_cutter.py
```diff
@@ -384,30 +384,42 @@ def quit_to_lobby(self):
...
 
     def simulate(self):
+        self.popup_watchdog = Clock.schedule_interval(lambda dt: self.set_simulation_popup_state(self.m.s.m_state), 1)
         if not self.is_config_valid():
             self.show_validation_popup()
             return
 
         if not self.simulation_started and self.m.s.m_state.lower() == 'idle':
-            self.m_state_bind = self.m.s.bind(m_state=lambda i, value: self.set_simulation_popup_state(value))
+            self.m.s.bind(m_state=lambda i, value: self.set_simulation_popup_state(value))
```
archiepeterjarvis, line 393: Happy to merge, but I think we should avoid using lambdas to hide parameters, especially in a bind. It makes it impossible to unbind, (you might not need to here, but I'm sure it will catch someone at some point). Also, keeping the signature of functions as intended by kivy is neater.

BenjiPestell: Understood, ta

PR #1715, opened 2024-04-29
Archie/unhandled exceptions

archiepeterjarvis:
# Send logs over pika on unhandled exception

- [ ] Ready for review

## Checklist
- [x] I have set the recent milestone
- [ ] I have updated the [change log](https://docs.google.com/document/d/1t3s6lQuUcug1Fj8E_i2gBJNvXN0-funwBArG2hsia6k/edit)

## Description
When the software runs into an unhandled exception, the log file is copied to 'crash.log' so that the software can detect that the machine crashed on the next boot. When it restarts, the user will be prompted as to whether they want to send the log file. If they choose to, it will be automatically sent over pika to the smart manager server and will be available to us & customer support.

## Notes
Does the serial number get output to the logs? If so, we need to remove it for GDPR, unless we're stating otherwise?

The text on the popup needs to be bulked out a bit.
image

### Visual Test
- [ ] Not applicable
- [ ] Own computer
- [x] Console 7"
- [x] Console 10" (needs scaling still)

### Function Test
- [ ] Not applicable
- [x] Own computer
- [x] Console 7"
- [x] Console 10"

### Unit Tests
- [ ] Not applicable
- [ ] Completed

## Screenshots (if applicable)

### DIFF
--- a/src/asmcnc/core_UI/new_popups/software_crashed_popup.py
+++ b/src/asmcnc/core_UI/new_popups/software_crashed_popup.py
@@ -50,7 +50,7 @@ class SoftwareCrashedPopup(PopupBase):
                 + " "
                   "Would you like to send a crash report to help diagnose the issue?"
         )
-        body = BoxLayout(size_hint_y=0.5, padding=[0, dp(10), 0, 0])
+        body = BoxLayout(size_hint_y=0.5, padding=scaling_utils.get_scaled_tuple([0, dp(10), 0, 0]))
         main_label = Label(text=main_string, font_size=scaling_utils.get_scaled_sp("15sp"),
                            color=color_provider.get_rgba("black"),
                            valign="top")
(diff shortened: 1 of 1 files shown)


src/asmcnc/core_UI/new_popups/software_crashed_popup.py
```diff
@@ -0,0 +1,124 @@
...
+CRASH_LOG = os.path.join(paths.COMMS_PATH, "logging_system", "logs", "crash.log")
+
+
+def check_for_crash():
+    return os.path.exists(CRASH_LOG)
+
+
+def serialize_log_file():
+    with open(CRASH_LOG, "rb") as log_file:
+        log_file_contents = log_file.read()
```
Masc999, line 28: We can put a regex replace here to remove serial numbers from the log.

PR #1752, opened 2024-06-12
Dwt file validation

DennisKiselev:
# Dwt file validation

[Jira Ticket](https://yetitoolsmartsoftware.atlassian.net/browse/YTSS-3009)
- [ ] Ready for review

## Checklist
- [ ] I have completed a self review
- [x] I have set the recent milestone
- [x] I have updated the jira ticket
- [ ] I have updated the [change log](https://docs.google.com/document/d/1t3s6lQuUcug1Fj8E_i2gBJNvXN0-funwBArG2hsia6k/edit)

## Description

Valid .dwt files have to contain a specific ID as a comment in the gcode. If it doesn't, the file is not loaded, and an error popup is shown.

### Visual Test
- [ ] Not applicable
- [x] Own computer
- [ ] Console 7"
- [ ] Console 10"

### Function Test
- [ ] Not applicable
- [x] Own computer
- [ ] Console 7"
- [ ] Console 10"

### Unit Tests
- [x] Not applicable
- [ ] Completed

## Screenshots (if applicable)

### DIFF
--- a/src/asmcnc/skavaUI/screen_file_loading.py
+++ b/src/asmcnc/skavaUI/screen_file_loading.py
@@ -27,6 +27,7 @@ from asmcnc.geometry import job_envelope
 from asmcnc.skavaUI import popup_info
 
 from asmcnc.core_UI.scaling_utils import get_scaled_sp
+from asmcnc.comms.model_manager import ModelManagerSingleton
 
 Builder.load_string(
     """
@@ -164,6 +165,7 @@ class LoadingScreen(Screen):
     default_font_size = get_scaled_sp("30sp")
     skip_check_decision = False
     continuing_to_recovery = False
+    dwt_file_id_line = "(DE590B1913285EFD678EFD0420ACD6B2)\n"
 
     def __init__(self, **kwargs):
         super(LoadingScreen, self).__init__(**kwargs)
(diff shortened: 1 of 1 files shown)


src/asmcnc/skavaUI/screen_file_loading.py
```diff
@@ -240,6 +243,11 @@ def objectifiled(self, job_file_path, dt):
             self.sm.pm.show_error_popup(file_empty_warning)
             self.sm.current = "local_filechooser"
             return
+        if self.model_manager.is_machine_drywall():
+            if not self.dwt_file_id_line in self.job_file_as_list:
```
BenjiPestell, line 247: Just a suggestion: In the interest of speed, can we limit the search to the first (x) number of lines maybe?

DennisKiselev: Is the ID going to be guaranteed to be at the top?

DennisKiselev: Also wanted to add, that since this line is only executed on drywall machines, on which only .dwt files are allowed to be loaded, this will only be slow in the case that somebody uses a non valid .dwt file, which should be rare?

BenjiPestell: Yeah that's reasonable

PR #1763, opened 2024-06-14
Archie/fix welcome

archiepeterjarvis:
# Fix language selection on welcome app

**On merging your PR, please copy the changelog to the master doc.**
- Header text now scrolls through languages (like previous version)
- Next button text now updates on language selection (like previous version)
- Changed file names to allow for testing on Windows (previously, the Korean/French/any with non EN characters wouldn't be found on Windows)
- Now uses paths module to find data consent .rst docs

## Checklist
- [ ] I have set the recent milestone
- [x] I have tested graphical changes on all languages
- [ ] I have added the relevant labels to this PR

## Description
This branch is based off of master, it may need to be rebased if it's going out with the hotfix. 

### Visual Test
- [ ] Not applicable
- [x] Own computer
- [x] Console 7"
- [ ] Console 10"

### Function Test
- [ ] Not applicable
- [x] Own computer
- [x] Console 7"
- [ ] Console 10"

### Unit Tests
- [ ] Not applicable
- [ ] Completed


LettyYeti (approved):
This is ok for merge as its better than what we have currently, but there are a couple of points that still need refinement: 

- The top banner is still different between the welcome screen and the following screen, so there's still a slight jump. It's much improved though. 
- Welcome text should "fix" to the selected language on selection
- Korean font doesn't work if you select one language, press Next, and then come back
- Fonts of the languages should be bigger

It's also worth mentioning that moving where the font for the following screens updates from has moved the lag that comes with it as well. Previously, there was lag on the language selection, now there's lag on pressing the Next button. I don't know which is better, so I've just made a ticket in Jira.

archiepeterjarvis:
Hi Lettie, yes there are some inconsistencies due to the methods used to build the previous version, which I'm trying to stay away from. In the previous version, the label itself had the height set, instead of the BoxLayout which makes it pretty painful for any future mods (e.g what if we wanted two labels).

It was also not at all functional on Windows which is what sparked the rewrite.

I will address your other points before merging as there's no rush on this until next release.

LettyYeti:
Good to know. 

On the label, the refactor is definitely a technical improvement, but obvs for the user the screen really needs to not jump around. 

For things like this going forwards, it's worth implementing that kind of change through the sequence (so just the banner adjustment in all start up screens) to ensure the smoothness of the UI is preserved. 

I've made a jira ticket for the other changes, I'd merge now anyway so that the improvements you've done so far are in asap

PR #1769, opened 2024-06-18
Add **kwargs to allow newer configs to run on older software

archiepeterjarvis:
# Add **kwargs to allow newer configs to run on older software

**On merging your PR, please copy the changelog to the master doc.**
- Add **kwargs to catch any parameters from newer DWT/Shapes config versions

## Checklist
- [ ] I have set the recent milestone

## Testing
I tested by adding tabs to my config, running the software then crashed. I then added kwargs to the config classes and it runs just fine. I also noticed the is_valid_configuration() func would trigger if the parameter count wasn't equal, so I've made it only try to fix configs with less parameters than expected. I tested this by removing the datum_position object from the config and running, ensuring that it catches the issue. 

** Note, there is a possible issue where the config has a new parameter, e.g. tabs but doesn't have an expected parameter, e.g. datum_position where the number of parameters would be correct, but the software would crash due to the wrong arguments provided. But this isn't really an expected real life situation (not at all sure how it would happen without manual tampering) **

### Visual Test
- [ ] Not applicable
- [ ] Own computer
- [ ] Console 7"
- [ ] Console 10"

### Function Test
- [ ] Not applicable
- [x] Own computer
- [ ] Console 7"
- [ ] Console 10"

### Unit Tests
- [ ] Not applicable
- [ ] Completed

### DIFF
--- a/src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
+++ b/src/asmcnc/apps/drywall_cutter_app/config/config_loader.py
@@ -164,7 +164,7 @@ class DWTConfig(EventDispatcher):
         field_count = len(cfg)
 
         valid_field_count = (
-                len(inspect.getargspec(config_classes.Configuration.__init__).args) - 1
+                len(inspect.getargspec(config_classes.Configuration.__init__).args[1:])
         )  # Subtract 1 for self (kwargs is not counted)
 
         # Config must have at least the same number of fields as the default config, but can have more (for future updates)


Masc999 (commented):
This has become somewhat obsolete, as Vlad is bringing together config and database and therefore changing over the config_classes. 
A new config with cutter_type = "0016" instead of "tool_6mm" will still crash. This should be caught when trying to load a config and checked against the configs version tag. If it is too new, the default config should be loaded.

PR #1771, opened 2024-06-19
Updated how the shape dimensions get checked

VladPavelescu:
# Update how shape dimensions get checked before running

[Jira Ticket](https://yetitoolsmartsoftware.atlassian.net/browse/YTSS-3047?atlOrigin=eyJpIjoiYjliMmRiYzhmOTgzNDUxZDg5MDlhODFkNTlhNWQxNDIiLCJwIjoiaiJ9)

Manually defocused all text inputs and manually ran the function which verifies that inputs are correct before a job is run.

## Checklist
- [X] I have completed a self review
- [X] I have set the recent milestone
- [X] I have updated the jira ticket
- [X] I have added the relevant labels to this PR

## Dependencies for merge
- [ ] [#<pull_request_number>]

### Visual Test
- [ ] Not applicable
- [X] Own computer
- [X] Console 7"
- [ ] Console 10"

### Function Test
- [ ] Not applicable
- [X] Own computer
- [X] Console 7"
- [ ] Console 10"

### Unit Tests
- [ ] Not applicable
- [ ] Completed

### DIFF
--- a/src/asmcnc/apps/drywall_cutter_app/widget_drywall_shape_display.py
+++ b/src/asmcnc/apps/drywall_cutter_app/widget_drywall_shape_display.py
@@ -4,6 +4,7 @@ from kivy.clock import Clock
 
 from asmcnc.apps.drywall_cutter_app.config.config_options import ToolpathOffsetOptions
 from asmcnc.comms.logging_system.logging_system import Logger
+from asmcnc.apps.drywall_cutter_app.config import config_options
 import re
 
 from asmcnc.core_UI import scaling_utils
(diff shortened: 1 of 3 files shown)


src/asmcnc/apps/drywall_cutter_app/widget_drywall_shape_display.py
```diff
@@ -552,7 +553,8 @@ def select_toolpath(self, shape, toolpath, rotation):
     def text_input_change(self, instance, *args):
         # On startup it seems to call this function and set everything to 0, so check that drywall app is open
         # Also check that the input matches a valid positive float, to allow user to correct mistakes
-        if self.sm.current == 'drywall_cutter' and re.match(r'^\d*\.\d+$', instance.text):
+        if ((self.dwt_config.app_type == config_options.AppType.SHAPES or
```
archiepeterjarvis, line 556: What other app type would it be? Is this really the same check as before?

VladPavelescu: For some reason, the check doesn't pass with the previous version of the code. If you don't click off the text input before pressing go, the new value doesn't get saved. With this, it does get saved. Again, not sure what causes this...

archiepeterjarvis: I'm confused why even check if app_type is SHAPES or DWT though? There is no other case. app_type isn't an indicator of the current app visible, only which machine version it is really.

archiepeterjarvis: I assume because go button takes you to a different screen and the sm has already updated? Not sure but I think you might be able to schedule the call for before the next frame which might solve it?

Masc999: It works fine the way it was. just tested it. The new code would only do something like "if True and ..."

Masc999 (approved):

